PR #6, opened 2020-04-22
Adding ability to format range of file

pergu:
Adding the ability to format a specific range of a file. In this iteration
 the function will either return a formatted string for the
  area that is being reformatted, or a list of alternate ranges
  which are ok, i.e. contains one or more forms.

In order to do this I'm also adding location annotations to raw_string in order to tell if the raw string is in the range that is being formatted.

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -86,10 +86,9 @@ format_range(FileName, StartLocation, EndLocation) ->
    try
         {ok, Forms, Warnings} = file_read_forms(FileName),
         case verify_ranges(Forms, StartLocation, EndLocation) of
-            ok ->
-                InRange = forms_in_range(Forms, StartLocation, EndLocation),
-                [$\n | Result] = format_forms(InRange),
-                verify_forms(FileName, InRange, Result),
+            {ok, FormsInRange} ->
+                [$\n | Result] = format_forms(FormsInRange),
+                verify_forms(FileName, FormsInRange, Result),
                 {ok, unicode:characters_to_binary(Result), Warnings};
             {options, Options} ->
                 {options, Options}
@@ -257,15 +256,17 @@ format_error(could_not_reparse) ->
     "formatter result invalid, could not reparse".
 
 verify_ranges(Forms, StartLocation, EndLocation) ->
-    case possible_ranges(Forms, StartLocation, EndLocation) of
+    ApplicableForms = forms_in_range(Forms, StartLocation, EndLocation),
+    case possible_ranges(ApplicableForms, StartLocation, EndLocation) of
         [{StartLocation, EndLocation}] ->
-            ok;
+            {ok, ApplicableForms};
         Options ->
             {options, Options}
     end.
 
+% Returns ranges which starts with the start of a form and ends with the end of form
 possible_ranges(Forms, StartLocation, EndLocation) ->
-    case forms_in_range(Forms, StartLocation, EndLocation) of
+    case Forms of
         [] ->
             [];
         [OnlyForm] ->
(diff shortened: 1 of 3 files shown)


src/erlfmt.erl
```diff
@@ -69,6 +74,30 @@ format_file(FileName, Out) ->
...
     end.
 
+-spec format_range(
+    file:name_all(),
+    erlfmt_scan:location(),
+    erlfmt_scan:location()
+) ->
+    {ok, string(), [error_info()]} |
+    {error, error_info()} |
+    {options, [{erlfmt_scan:location(), erlfmt_scan:location()}]}.
```
michalmuskala, line 84: I wonder - if we expose this operation through CLI, what do you think should be the behaviour in case there are multiple options?

src/erlfmt.erl
```diff
@@ -69,6 +74,30 @@ format_file(FileName, Out) ->
...
+format_range(FileName, StartLocation, EndLocation) ->
+   try
+        {ok, Forms, Warnings} = file_read_forms(FileName),
+        case verify_ranges(Forms, StartLocation, EndLocation) of
+            ok ->
+                InRange = forms_in_range(Forms, StartLocation, EndLocation),
```
michalmuskala, line 90: It looks like `verify_ranges` ends up calling `forms_in_range` as well. Would it make sense to perform this computation only once?

src/erlfmt.erl
```diff
@@ -220,3 +255,51 @@ format_error({not_equivalent, Node1, Node2}) ->
...
+possible_ranges(Forms, StartLocation, EndLocation) ->
+    case forms_in_range(Forms, StartLocation, EndLocation) of
+        [] ->
+            [];
+        [OnlyForm] ->
+            [get_location_range(OnlyForm)];
+        MultipleForms ->
+            combine(
+                get_possible_locations(MultipleForms, StartLocation, fun get_location/1),
+                get_possible_locations(lists:reverse(MultipleForms), EndLocation, fun get_end_location/1))
```
michalmuskala, line 276: I'm not sure I understand correctly what is happening here. For successful operation we only accept ranges that overlap exactly one form, but when there are multiple overlapping forms we return as options all the possible ranges overlapping all of them. Is that correct? If we only accept ranges with one form, here we should return only ranges that contain one form.

pergu: So we allow ranges that exactly contains 1 or more form, so the options are the possible forms you could have meant to include say for example that you have a range with three forms, where the range starts in form 1, form 2 is fully contained in the range and form 3 ends after the range so you have two sensible start locations the start of form 1 and the start of form 2, and two sensible end locations the end of form 2 and the end of form 3.
I.e. you end up with four possible ranges {1,2,3}, {2,3}, {1,2} and {2}.

src/erlfmt.erl
pergu: Yes, I can move that in so that verify_ranges also returns the range.

src/erlfmt.erl
pergu: Yeah I guess either there could be an interactive choice, or there is a default way to resolve these choice that can be altered with command line flags. I.e. the default is always choose the min range but you could choose to include a half baked form if it precedes or the range and same at the end.

src/erlfmt.erl
pergu: Pushed a new version where this is not done twice.

slaykachu:
Apparently, the endpoint in not inclusive.
If I use {1,1}, {1,1} as a range, the function returns an empty `PossibleRanges` list.
Was that on purpose?

PR #7, opened 2020-04-23
Preserve empty lines between comments

michalmuskala:
This modifies how comments are printed back, preserving empty lines
between comment groups. Before, they would all be squashed into one
sequence.

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -214,6 +214,9 @@ combine_sep(D1, Sep, D2) ->
 combine_newline(D1, D2) ->
     document_combine(document_flush(D1), D2).
 
+combine_double_newline(D1, D2) ->
+    document_combine(document_flush(document_flush(D1)), D2).
+
 combine_comma_newline(D1, D2) ->
     Left = document_flush(document_combine(D1, document_text(","))),
     document_combine(Left, D2).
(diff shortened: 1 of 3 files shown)


test/erlfmt_SUITE_data/comments.erl.formatted
```diff
@@ -3,6 +3,7 @@
 -export([has_fanciness/1, bar/1, baz/1]).
 
 %% Constants
+
```
pergu, line 6: This behaviour seems counter-intuitive to me i.e.:

%% commmentcommentcomment
long line of thing... % comment

becoming:

%% commmentcommentcomment

% comment
long line of thing...

michalmuskala: Yeah, I wasn't 100% sure about this. What would be the desired behaviour? Something like this?

```
%% commmentcommentcomment
% comment
long line of thing...
```

pergu: Yeah, that would be my intuition

michalmuskala:
Updated according to the suggestions

michalmuskala:
Alternative implementation in https://github.com/WhatsApp/erlfmt/pull/35

PR #11, opened 2020-05-05
A test that checks that markdown files in the repo root and doc folder are formatted with erlfmt

awalterschulze:


test/erlfmt_markdown_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+
+markdown_files(Config) when is_list(Config) ->
+    {ok, Data} = file:get_cwd(),
+    RepoRootPath = filename:join([Data, "..", "..", "..", ".."]),
+    DocPath = filename:join(RepoRootPath, "doc"),
+    Filenames = lists:append(
+        find_markdown_filenames_in(RepoRootPath),
+        find_markdown_filenames_in(DocPath)
+    ),
+    ?assertMatch([_One | [_Two | [_Three | _AtLeast]]], Filenames),
```
michalmuskala, line 79: You can write this as
```suggestion
    ?assertMatch([_One, _Two, _Three | _AtLeast], Filenames),
```

test/erlfmt_markdown_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+            check_markdown(binary_to_list(Content))
+        end,
+        Filenames
+    ).
+
+find_markdown_filenames_in(Path) ->
+    {ok, BaseFilenames} = file:list_dir_all(Path),
+    Filenames =
+        lists:map(fun (Filename) -> filename:join([Path, Filename]) end, BaseFilenames),
+    lists:filter(fun (Filename) -> filename:extension(Filename) == ".md" end, Filenames).
```
michalmuskala, line 92: I think this logic could be replaced with `filelib:wildcard/1`.

test/erlfmt_markdown_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+        _:_ ->
+            throw("Big Problem Parsing: " ++ String)
+    end.
+
+format_doc(String) ->
+    case erlfmt:read_forms_string("nofile", "f() ->\n" ++ String ++ ".") of
+        {ok, [{function, _, [{clause, _, _, empty, [Expr]}]}], []} ->
+            erlfmt_format:expr_to_algebra(Expr);
+        _ ->
+            {ok, [Form], []} = erlfmt:read_forms_string("nofile", String),
```
michalmuskala, line 173: I think this is something we'll need to do in general to support formatting files like `rebar.config`, but :+1: for now

README.md
```diff
@@ -12,20 +12,20 @@ fits within the selected maximum line-length.
 
 For example, the following snippet:
 
-```erl
+```erl formatted hello
```
michalmuskala, line 15: I really like how this works! And I haven't thought about checking readme either. Good job!

test/erlfmt_markdown_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+                                    check_fmt(PreviousCode, CurrentCode);
+                                "unformatted" ->
+                                    check_fmt(CurrentCode, PreviousCode)
+                            end,
+                            {text, maps:remove(Key, State)}
+                    end
+            end;
+        _ ->
+            {text, State}
+    end.
```
michalmuskala, line 152: I think this could be clearer if the two phases - gathering chunks to compare and comparison itself were separated.
We could accumulate all the examples into a map like:
```erl
#{Key => {formatted, Formatted} | {unformatted, Unformatted} | {check, Formatted, Unformatted}}
```
 and then at the end traversing the map. This would also allow detecting if keys have typos in them and things are not checked or are accidentally duplicated.

test/erlfmt_markdown_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+    Sections = string:split(Content, "```", all),
+    lists:foldl(fun check_section/2, {text, maps:new()}, Sections).
+
+check_section(_Text, {text, State}) ->
+    {code, State};
+check_section(Text, {code, State}) ->
+    [FirstLine, CurrentCode] = string:split(Text, "\n"),
+    Spec = string:split(FirstLine, " ", all),
+    case Spec of
+        [Erlang | [ToFormat | Keys]] ->
```
michalmuskala, line 125: This could be simplified and combined with the following prefix check using something like:
```erl
["erl" ++ _, ToFormat | Keys]
```

test/erlfmt_markdown_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+            case string:prefix(Erlang, "erl") of
+                nomatch ->
+                    {text, State};
+                _ ->
+                    Key = string:join(Keys, " "),
```
michalmuskala, line 130: I guess this is not strictly needed, the map key could be a list of strings. It doesn't make much of a difference, though

test/erlfmt_markdown_SUITE.erl
```diff
@@ -0,0 +1,175 @@
...
+                            {text, maps:remove(Key, State)}
+                    end
+            end;
+        _ ->
+            {text, State}
+    end.
+
+check_fmt(Unformatted, Expected) ->
+    NewlyFormatted = format(Unformatted, 80),
+    ?assertEqual(string:trim(NewlyFormatted), string:trim(Expected)).
```
michalmuskala, line 156: If you go for the suggestion about changing how checking works, would also add here a check that formatting is idempotent - that the expected string after formatting is equal to itself. I think right now this only happens in the loop above if we first hit a formatted example followed by an unformatted one, in the reverse case it's not checked.

test/erlfmt_markdown_SUITE.erl
awalterschulze: Yes, even this function failed to format

```erl
{plugins, [erlfmt]}.
```

and

```erl
{erlfmt, [
    write,
    {files, "{src,include,test}/*.{hrl,erl}"}
]}.
```

But that would be nice to add in a follow up commit.

README.md
awalterschulze: It turns out the use case I needed was already there :)
I also stumbled upon this and decided to use this and then we can add more markdown files in doc and root and they will automatically be checked from the get go.

test/erlfmt_markdown_SUITE.erl
awalterschulze: I was looking for this, thank you

test/erlfmt_markdown_SUITE.erl
awalterschulze: I forgot how to do this, thank you.

test/erlfmt_markdown_SUITE.erl
awalterschulze: nice :)

test/erlfmt_markdown_SUITE.erl
awalterschulze: Good catch :)

test/erlfmt_markdown_SUITE.erl
awalterschulze: Great idea.

awalterschulze:
I think this is ready to review again.

test/erlfmt_markdown_SUITE.erl
```diff
@@ -0,0 +1,180 @@
...
+        maps:size(
+            maps:fold(
+                fun (Key, _Value, FriendsLeft) ->
+                    maps:remove(Key, FriendsLeft)
+                end,
+                Unformatted,
+                Formatted
+            )
+        )
+    ).
```
michalmuskala, line 140: I would probably write this as
```suggestion
    ?assertEqual(#{}, maps:without(maps:keys(Formatted), Unformatted)).
```

awalterschulze: Thank you this is great.
I am learning so much from this review.
Thank you for taking the time.

michalmuskala:
:shipit:

PR #27, opened 2020-05-20
Support parsing standalone expressions 

### LINKED ISSUE #13: Format files such as rebar.config or sys.config

michalmuskala:
This unblocks formatting files like `rebar.config`, but also simplifies
testing by unifying forms and expressions. We support both a sequence of
dot-terminated expressions as well as a single expression without a dot.

Additionally, since we can easily parse both forms and standalone expressions, there's
no need for two sets of macros in `erlfmt_format_SUITE`.

Closes #13

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -15,7 +15,7 @@
 
 -include("erlfmt.hrl").
 
--export([form_to_algebra/1]).
+-export([to_algebra/1]).
 
 -import(erlfmt_algebra, [
     document_text/1,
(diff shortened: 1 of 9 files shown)


src/erlfmt_format.erl
```diff
@@ -81,7 +81,16 @@ form_to_algebra({attribute, Meta, {atom, _, record}, [Name, {tuple, TMeta, Value
 form_to_algebra({attribute, Meta, Name, Values}) ->
     Prefix = wrap(document_text("-"), expr_to_algebra(Name), document_text("(")),
     Doc = container_to_algebra(Meta, Values, Prefix, document_text(").")),
-    combine_comments(Meta, Doc).
+    combine_comments(Meta, Doc);
+form_to_algebra(Expr) ->
```
awalterschulze, line 85: Maybe this function name is now just `to_algebra` or `new_algebra`?

src/erlfmt_parse.yrl
```diff
@@ -1014,7 +1016,15 @@ parse_form([{'-',A1},{atom,A2,define}|Tokens]) ->
             parse(NewTokens3)
     end;
 parse_form(Tokens) ->
-    parse(Tokens).
+    case parse(Tokens) of
+        {ok, _} = Res ->
+            Res;
+        Error ->
+            case parse([{standalone_expr, #{}} | Tokens]) of
```
awalterschulze, line 1023: I would create a function parse_expr that does this and then a function parse_any that tries parse_form and then parse_expr, since the name `parse` is already taken.  Now I find it slightly misleading, since parse_form implies that it will only parse a form, but maybe I am missing something?

test/erlfmt_format_SUITE.erl
```diff
@@ -550,17 +537,17 @@ binary(Config) when is_list(Config) ->
     ).
 
 map_create(Config) when is_list(Config) ->
-    ?assertFormatExpr("#{\n}", "#{}"),
-    ?assertFormatExpr("#{1:=2,  3=>4}", "#{1 := 2, 3 => 4}"),
-    ?assertFormatExpr(
+    ?assertFormat("#{\n}", "#{}"),
```
awalterschulze, line 540: Really like that assertFormatExpr is now just assertFormat

michalmuskala:
Yeah, I agree at this point the distinction between `form` and `expression` is a bit meaningless. I'll remove it from more places

michalmuskala:
I switched places that used to handle just forms from `form` to `node` naming. This should make it clearer that it can be either form or expression.

awalterschulze:
This looks great :)

PR #28, opened 2020-05-21
--require-pragma

awalterschulze:


src/erlfmt.erl
```diff
@@ -22,15 +22,18 @@
...
 ]).
 
--export_type([error_info/0, out/0]).
+-export_type([error_info/0, out/0, config/0]).
 
 -type error_info() :: {file:name_all(), erl_anno:location(), module(), Reason :: any()}.
 
 -type out() :: standard_out | {path, file:name_all()} | replace.
 
+-type config() :: {RequirePragma :: boolean(), Out :: out()}.
```
awalterschulze, line 35: This config type is created to make space for InsertPragma in a following pull request

test/erlfmt_pragma_SUITE.erl
```diff
@@ -0,0 +1,124 @@
+%% Copyright (c) Facebook, Inc. and its affiliates.
```
awalterschulze, line 1: Insert pragma tests will also go into this file.

src/erlfmt.erl
```diff
@@ -64,18 +67,42 @@ init(State) ->
...
+        _ -> false
+        end.
+
+contains_pragma_comment({comment, _Loc, Comments}) ->
+    lists:any(fun (Comment) ->
+        case string:find(Comment, "@format") of
+            nomatch -> false;
+            _ -> true
+        end
+    end, Comments);
```
michalmuskala, line 103: A list of strings is also a string, so this can be simplified to 
```suggestion
    string:find(Comments, "@format") =/= nomatch;
```

test/erlfmt_pragma_SUITE.erl
michalmuskala: I'd avoid creating a separate suite for this. In general performance of CT scales really poorly with a lot of small suites (they can't run in parallel). I think making a group in `erlfmt_SUITE` for those would be enough

src/erlfmt.erl
```diff
@@ -64,18 +67,42 @@ init(State) ->
...
-format_file(FileName, Out) ->
+-spec format_file(file:name_all(), config()) -> {ok, [error_info()]} | {error, error_info()}.
+format_file(FileName, {RequirePragma, Out}) ->
     try
         {ok, Forms, Warnings} = file_read_forms(FileName),
-        [$\n | Formatted] = format_forms(Forms),
-        verify_forms(FileName, Forms, Formatted),
-        write_forms(FileName, Formatted, Out),
-        {ok, Warnings}
+        ShouldFormat = not RequirePragma or contains_pragma_forms(Forms),
```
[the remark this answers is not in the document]
awalterschulze: Yes that makes sense, or at least parse until some code appears.
I think it is quite natural to want to place it after some licensing comments.

src/erlfmt.erl
awalterschulze: Wait what?
`string:find` can search a list of comments?
Oh because a list of strings, is just a list of characters and a list of list of characters, is flattened like an IO list, when finding?

test/erlfmt_pragma_SUITE.erl
awalterschulze: Hmmm, that is interesting.
I am worried that erlfmt_SUITE is already over 900 lines long.
It is starting to become hard to read and find things.

src/erlfmt.erl
awalterschulze: Makes sense, took me a while to see:

no require, no insert => current
no require, insert => format all files, but insert pragma after formatting.
require, no insert => only format files with pragma
insert, require => Ah, I see this option doesn't make sense

src/erlfmt.erl
michalmuskala: Yeah, it's because of a way `string` module defines strings as `unicode:chardata()` which expands to something like:
```erl
chardata() :: unicode_binary() | charlist().
charlist() :: maybe_improper_list(char() | unicode_binary() | charlist(), unicode_binary() | []).
```

test/erlfmt_pragma_SUITE.erl
michalmuskala: I agree. I think it would make sense to extract the parser tests from `erlfmt_SUITE` into a separate file and leave `erlfmt_SUITE` for the more integration-style tests. This can be done in a new PR

awalterschulze: Oh okay, yes that can work :)

src/erlfmt.erl
awalterschulze: Who needs monads when you have the type doing the flattening for you :)

```
result(v, e) :: ok string | error(e)
error(e) :: maybe_improper_list(e)
```

Thank you this is really good to know

src/erlfmt.erl
```diff
@@ -64,18 +69,35 @@ init(State) ->
...
-format_file(FileName, Out) ->
+-spec format_file(file:name_all(), config()) -> {ok, [error_info()]} | {error, error_info()}.
+format_file(FileName, {Pragma, Out}) ->
     try
         {ok, Forms, Warnings} = file_read_forms(FileName),
-        [$\n | Formatted] = format_forms(Forms),
-        verify_forms(FileName, Forms, Formatted),
-        write_forms(FileName, Formatted, Out),
-        {ok, Warnings}
+        ShouldFormat = (Pragma == ignore) or contains_pragma_forms(Forms),
```
michalmuskala, line 76: This is good progress, but it still parses the whole file just to check the first couple lines. Since in `file_read_forms` we special-case the first form already (by calling `io_scan` before entering the main loop), I was thinking we could check this in there and skip parsing the rest of the file if we don't detect the pragma. I think it would be reasonable to return `ignore` from this function if that happens.

michalmuskala: As an addition, please avoid using `and` and `or` in favour of `andalso` and `orelse`. The short forms are eager, so `or` evaluates the right side even if the left side is false.

awalterschulze: Oh `orelse ` that is very interesting.

Ah yes, `file_read_forms`, great idea.

awalterschulze: These are both addressed in newest commit, I hope.

awalterschulze:
A squash and merge commit will be required, since I merged in master.

PR #30, opened 2020-06-14
Remove warning about # when formatting escript

ebengt:
The check for removing the warning can be considered as too much/brittle. Maybe it is not necessary to check for ''escript'', and instead always remove the warning about # first in the file.

As can be seen in the escript that is formatted, it is worth investigating why the first line formatted is not correctly formatted. It is still
main( [_] ) ->
after formatting.
Current erlfmt will remove those extra spaces, except when formatting escript.
This is not due to this change.

Finally, I have not formatted erlfmt.erl with erlfmt. That creates a lot of changes in code that I have not touched.

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -181,9 +181,24 @@ parse_nodes(Tokens, Comments, FileName, Cont, Warnings) ->
         {ok, Node} ->
             {erlfmt_recomment:recomment(Node, Comments), Warnings};
         {error, {ErrLoc, Mod, Reason}} ->
-            Warning = {FileName, ErrLoc, Mod, Reason},
-            {node_string(Cont), [Warning | Warnings]}
+            New_warnings =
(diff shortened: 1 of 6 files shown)


ebengt:
Dialyzer complains about
The pattern <Warning = {_, #{'location':={1, 1}, 'end_location':={1, 2}}, 'erlfmt_parse', ["syntax error before: ", "'#'"]}, Tokens, Warnings> can never match the type
The match occurs, so it is perhaps a problem with the second type?
erl_anno:line() is  line() = integer() >= 0
Perhaps it is actually erl_anno:anno() instead?
On my system (22.2.5) there is no complaint from Dialyzer, so I have not tested this theory (yet).

michalmuskala:
I pushed in https://github.com/WhatsApp/erlfmt/commit/6e6f98c8011aedef0ee8786838f6631adf65d003 and alternative fix that tries to pre-read the shebang before parsing the rest of the file instead of rescuing the errors.

ebengt:
Pre-reading is a better solution. Thank you.

PR #32, opened 2020-06-23
add formatting check to rebar3 config in readme, since it is now supported

awalterschulze:
This is what I was trying to say with the comment, here https://github.com/WhatsApp/erlfmt/issues/13#issuecomment-632230417


michalmuskala:
Thank you! :heart:

PR #35, opened 2020-07-02
Switch to a greedy layout algorithm

### LINKED ISSUE #16: Large binaries and tuples take forever to format
This is caused by `erlfmt_format:fill_container_to_algebra/4` having exponential complexity. The simplest solution would probably be to introduce a new formatting primitive for fill layouts. This should be possible since they are flexible and adjust to fill the remaining space, so don't affect the general space calculation used by metrics.

### LINKED ISSUE #20: Revisit placement of when
The general decision was for `when` to be placed either inline (when the guard is single-line) or on a new line doubly-indented. This worked well, except for the case of extremely small functions and multiline guards (for example, because of a comment):

```erl
bar(X)
        when is_tuple(X);
             % some comment
             is_list(X) ->
    ...
```

because of this, I decided to allow "floating" guards. With that change, the example above would look like this:

```erl
bar(X) when is_tuple(X);
            % some comment
            is_list(X) ->
    ...
```

Which looks much better, but has the unfortunate effect of "floating" guards that need to be moved when the function name changes - this is generally something we'd really want to avoid.

Some possible solutions are:
* don't indent `when`. This creates relatively little differentiation between guards and body of the function
```erl
bar(X)
when is_tuple(X);
     % some comment
     is_list(X) ->
    ...
```

* indent `when` half the indentation - this is better, but rather awkward - we don't use half-indentation anywhere else
```erl
bar(X)
  when is_tuple(X);
       % some comment
       is_list(X) ->
    ...
```

* indent `when` full indentation - this is good in for multiline guards, but fails the guard-body distinction check when guard is on a separate line, but single-line
```erl
bar(Some, Very, Long, Arguments)
    when is_tuple(X); is_list(X) ->
    ...
```

* treat `when` and `->` as a sort-of parenthesis - this preserves clarity, but has relatively low current usage
```erl
bar(X) when 
    is_tuple(X);
    % some comment
    is_list(X) 
->
    ...
```

### LINKED ISSUE #24: Preserve empty line between comment and form
We should preserve empty line between comment and the expression it's attached to. 

Example use case is a license header followed by a newline before the rest of the code.

michalmuskala:
This migrates to a slightly modified version of the algebra defined in the ["Strictly Pretty" (2000) by Christian Lindig][0] as modified and implemented in [Elixir][1].

The primary issue with the previous algebra was performance of "flex" layouts as hit in https://github.com/WhatsApp/erlfmt/issues/16.

Fixes #16 
Fixes #20 
Fixes #24 

Huge thanks to @awalterschulze for pairing on this one with me!

[0]: http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.34.2200
[1]: https://github.com/elixir-lang/elixir/blob/f83724602773cea85445aae7ad25615793057e46/lib/elixir/lib/inspect/algebra.ex


test/erlfmt_format_SUITE.erl
```diff
@@ -254,7 +263,7 @@ unary_operator(Config) when is_list(Config) ->
     ?assertFormat("(bnot 1)*1", "(bnot 1) * 1"),
     ?assertSame("(catch 1) + 1"),
 
-    %% Unless it's nested not or bnot
+    % %% Unless it's nested not or bnot
```
awalterschulze, line 266:
```suggestion
    %% Unless it's nested not or bnot
```

test/erlfmt_format_SUITE.erl
```diff
@@ -231,19 +231,28 @@ string_concat(Config) when is_list(Config) ->
...
     ?assertFormat("- 1", "-1"),
 
-    %% Formats word operators with space
+    % %% Formats word operators with space
     ?assertSame("bnot 1"),
     ?assertSame("not true"),
     ?assertSame("catch 1"),
 
-    %% Does not change parenthesis, only whitespace
+    % %% Does not change parenthesis, only whitespace
```
awalterschulze, line 255:
```suggestion
    %% Does not change parenthesis, only whitespace
```

test/erlfmt_format_SUITE.erl
```diff
@@ -231,19 +231,28 @@ string_concat(Config) when is_list(Config) ->
...
+        ")"
     ).
 
 unary_operator(Config) when is_list(Config) ->
     %% Formats symbolic operators without space
     ?assertFormat("+ 1", "+1"),
     ?assertFormat("- 1", "-1"),
 
-    %% Formats word operators with space
+    % %% Formats word operators with space
```
awalterschulze, line 250:
```suggestion
    %% Formats word operators with space
```

PR #37, opened 2020-07-03
erlfmt: remove empty lines below single line attributes

### LINKED ISSUE #21: Empty lines between preprocessor directives
Hi Michal

Thanks a lot for the great work so far!

Let's take this file, not yet formatted:

```
-module(foo).
-inlude("foo.hrl").
-inlude("bar.hrl").
-export([foo/0]).
-export([bar/0]).
-define(FOO, "foo").
-define(BAR, "bar").
foo() -> ?FOO.
bar() -> ?BAR.
```
After formating, it becomes:

```
-module(foo).

-inlude("foo.hrl").

-inlude("bar.hrl").

-export([foo/0]).

-export([bar/0]).

-define(FOO, "foo").

-define(BAR, "bar").

foo() -> ?FOO.

bar() -> ?BAR.
```

It would be great to include an empty line only if the next directive is different, in order to group similar directives. This would produce the following output which I find nicer. But it's a matter of taste. ;)

```
-module(foo).

-inlude("foo.hrl").
-inlude("bar.hrl").

-export([foo/0]).
-export([bar/0]).

-define(FOO, "foo").
-define(BAR, "bar").

foo() -> ?FOO.

bar() -> ?BAR.
```
What do you think?

Nicolas

awalterschulze:
Fixes #21

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -196,17 +196,40 @@ node_string(Cont) ->
     {raw_string, Anno, string:trim(String, both, "\n")}.
 
 format_nodes([{attribute, _, {atom, _, spec}, _} = Attr, {function, _, _} = Fun | Rest]) ->
-    [$\n, format_node(Attr), format_node(Fun) | format_nodes(Rest)];
+    [$\n, format_node(Attr), $\n, format_node(Fun), $\n | format_nodes(Rest)];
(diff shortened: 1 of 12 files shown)


ntalfer:
thanks @awalterschulze !

awalterschulze:
and thanks @michalmuskala 
It is not clear from the commit, but we pair programmed on this :)

PR #60, opened 2020-07-13
a first draft of a formatting decision document for comments

awalterschulze:


doc/FormattingDecisionComments.md
```diff
@@ -0,0 +1,201 @@
...
+
+### Erlang.se
+
+[Programming Rules and Conventions, Section 8.5 Comments](http://www.erlang.se/doc/programming_rules.shtml#HDR36)
+
+```
+  1. Comments within Erlang code shall start with one percent character (%). If a line only contains a comment, it shall be indented as Erlang code. This kind of comment shall be placed above the statement it refers to. If the comment can be placed at the same line as the statement, this is preferred.
+  2. Comments about a function shall be without indentation and start with two percent characters (%%)
+  3. Comments about the module shall be without indentation and start with three percent characters (%%%)
+```
```
michalmuskala, line 15: Let's make this section into a quote instead of a block. It line-wraps nicer

```suggestion
> 1. Comments within Erlang code shall start with one percent character (%). If a line only contains a comment, it shall be indented as Erlang code. This kind of comment shall be placed above the statement it refers to. If the comment can be placed at the same line as the statement, this is preferred.
> 2. Comments about a function shall be without indentation and start with two percent characters (%%)
> 3. Comments about the module shall be without indentation and start with three percent characters (%%%)
```

doc/FormattingDecisionComments.md
```diff
@@ -0,0 +1,201 @@
...
+
+```
+Style-wise, comments that follow code on the same line are usually written with only a single % character, whereas comments that are on lines of their own are typically written starting with two % characters ...
+
+(Some people even like to start with three % characters on comment lines that describe things on a whole-file level, such as comments at the top of the source file.)
+
+One good reason to stick to these conventions is that syntax-aware editors such as Emacs and erlIDE can be made to know about them, so that they will indent comments automatically according to how many % characters they begin with.
+
+Now that you have a source file that defines a module, you need to compile it.
+```
```
michalmuskala, line 28: Similar here

```suggestion
> Style-wise, comments that follow code on the same line are usually written with only a single % character, whereas comments that are on lines of their own are typically written starting with two % characters ...
>
> (Some people even like to start with three % characters on comment lines that describe things on a whole-file level, such as comments at the top of the source file.)
>
> One good reason to stick to these conventions is that syntax-aware editors such as Emacs and erlIDE can be made to know about them, so that they will indent comments automatically according to how many % characters they begin with.
>
> Now that you have a source file that defines a module, you need to compile it.
```

michalmuskala:
Should we merge this?

michalmuskala:
@awalterschulze ping

awalterschulze:
I am still waiting for feedback from Robert, but don't want to push this, as he is on vacation.

awalterschulze:
We can still comment on this after it is merged.  So going to merge for now.

PR #63, opened 2020-07-15
--insert-pragma inserts an @format comment into a file

### LINKED ISSUE #15: Add @format pragma handling
We should add a way to adopt formatter gradually on a set of files. 
This would be simplest done requiring the first comment block to contain a `@format` annotation.  Please see https://prettier.io/docs/en/cli.html#--require-pragma for an example.

Additionally add a `--insert-pragma` command that formats a file and adds a pragma for easier step-by-step adoption. Please see https://prettier.io/docs/en/cli.html#--insert-pragma for an example.

awalterschulze:
Fixes #15

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -24,7 +24,6 @@
     format_error/1,
     format_error_info/1,
     format_nodes/1,
-    contains_pragma_file/1,
     contains_pragma_string/2,
     insert_pragma_nodes/1
 ]).
@@ -86,44 +85,36 @@ format_file(FileName, {Pragma, Out}) ->
         {error, Error} -> {error, Error}
     end.
 
--spec contains_pragma_file(file:name_all()) -> boolean().
-contains_pragma_file(FileName) ->
-    read_file(FileName, fun (File) ->
-        read_pragma_nodes(erlfmt_scan:io_node(File), FileName, [], [])
-    end).
-
--spec contains_pragma_string(file:name_all(), string()) -> boolean().
+-spec contains_pragma_string(file:name_all(), string()) -> boolean() | {error, error_info()}.
 contains_pragma_string(FileName, String) ->
-    read_pragma_nodes(erlfmt_scan:string_node(String), FileName, [], []).
+    try read_nodes(erlfmt_scan:string_node(String), FileName, require, [], [], []) of
+        {skip, _} -> false;
+        _ -> true
+    catch
+        {error, Error} -> {error, Error}
+    end.
 
-read_pragma_nodes({ok, Tokens, Comments, Cont}, FileName, _Acc, Warnings0) ->
-    {Node, _Warnings} = parse_nodes(Tokens, Comments, FileName, Cont, Warnings0),
-    contains_pragma_node(Node);
-read_pragma_nodes(_, _, _, _) ->
-    false.
+is_shebang_node({raw_string, _, "#!" ++ _}) -> true;
+is_shebang_node(_) -> false.
 
-contains_pragma_node({attribute, Meta, _AfAtom, _AbstractExprs}) ->
-    {PreComments, PostComments} = erlfmt_format:comments(Meta),
-    lists:any(fun contains_pragma_comment/1, PreComments ++ PostComments);
-contains_pragma_node(_) ->
-    false.
+contains_pragma_node(Node) ->
+    {PreComments, PostComments} = erlfmt_format:comments(Node),
+    lists:any(fun contains_pragma_comment/1, PreComments ++ PostComments).
 
 %% insert_pragma_nodes only inserts an @format comment,
 %% if one has not already been inserted.
 insert_pragma_nodes([]) -> [];
 insert_pragma_nodes([Node | Nodes]) ->
-    case contains_pragma_node(Node) of
-        true -> [Node | Nodes];
-        false -> [insert_pragma_node(Node) | Nodes]
+    case is_shebang_node(Node) of
+        true -> [Node | insert_pragma_nodes(Nodes)];
+        false -> case contains_pragma_node(Node) of
+            true -> [Node | Nodes];
+            false -> [insert_pragma_node(Node) | Nodes]
+        end
     end.
 
-insert_pragma_node({attribute, Meta, AfAtom, AbstractExprs}) ->
-    {attribute, insert_pragma_meta(Meta), AfAtom, AbstractExprs};
 insert_pragma_node(Node) ->
-    Node.
-
-insert_pragma_meta(Meta) ->
-    PreComments = erlfmt_scan:get_anno(pre_comments, Meta, []),
+    PreComments = erlfmt_scan:get_anno(pre_comments, Node, []),
     NewPreComments =
         case PreComments of
             [] -> [{comment, #{end_location => {2,1},location => {1,1}}, ["%% @format", ""]}];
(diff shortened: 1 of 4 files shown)


src/erlfmt.erl
```diff
@@ -102,6 +108,31 @@ contains_pragma_node({attribute, Meta, _AfAtom, _AbstractExprs}) ->
...
+%% insert_pragma_nodes only inserts an @format comment,
+%% if one has not already been inserted.
+insert_pragma_nodes([]) -> [];
+insert_pragma_nodes([Node | Nodes]) ->
+    case contains_pragma_node(Node) of
+        true -> [Node | Nodes];
+        false -> [insert_pragma_node(Node) | Nodes]
+    end.
+
+insert_pragma_node({attribute, Meta, AfAtom, AbstractExprs}) ->
```
michalmuskala, line 120: It's not necessarily an attribute as the first node. It could be any expression in something like `rebar.config`. Using `erlfmt_scan:get_anno` and `erlfmt_scan:put_anno` directly on the `Node` should work just fine - there's no need for extracting the `Meta` part manually.

test/erlfmt_SUITE.erl
```diff
@@ -1106,3 +1135,95 @@ contains_pragma(Config) when is_list(Config) ->
...
+    StringWithPragma.
+
+format_string(String, Pragma) ->
+    {ok, Nodes, []} = erlfmt:read_nodes_string("nofile", String),
+    NodesWithPragma = case Pragma of
+        insert -> erlfmt:insert_pragma_nodes(Nodes);
+        _ -> Nodes
+    end,
+    [$\n | Formatted] = erlfmt:format_nodes(NodesWithPragma),
+    unicode:characters_to_list(Formatted).
```
michalmuskala, line 1229: I'm not convinced about these unit tests. What do you think about adding a `erlfmt:format_string` function that would allow us to avoid exposing the two internal `erlfmt:insert_pragma_nodes` and `erlfmt:format_nodes` functions? It should allow to avoid increasing the public API area and simplifying other tests in turn as well where we're repeating a similar pattern.

test/erlfmt_SUITE.erl
```diff
@@ -1106,3 +1135,95 @@ contains_pragma(Config) when is_list(Config) ->
...
+        insert_pragma_string(
+            "%% LICENSE\n"
+            "%% LICENSE\n"
+            "%% LICENSE\n"
+            "%% LICENSE\n"
+            "\n"
+            "-module(pragma)\n."
+            "\n"
+            "-export([f/3]).\n"
+        )
```
michalmuskala, line 1213: This seems to be formatted funny. Could you erlfmt the file? 😅

test/erlfmt_SUITE.erl
```diff
@@ -956,6 +964,27 @@ smoke_test_stdio_with_pragma(Config) when is_list(Config) ->
...
+    Path = filename:join(DataDir, "pragma.erl"),
+    Formatted = os:cmd("cat " ++ Path ++ " | " ++ escript() ++ " - --insert-pragma"),
+    {ok, Expected} = file:read_file(Path),
+    ?assertEqual(Expected, unicode:characters_to_binary(Formatted)).
+
+smoke_test_stdio_insert_and_require_pragma(Config) when is_list(Config) ->
+    DataDir = ?config(data_dir, Config),
+    Path = filename:join(DataDir, "pragma.erl"),
+    ErrorString = os:cmd("cat " ++ Path ++ " | " ++ escript() ++ " - --insert-pragma --require-pragma"),
+    ?assert(string:find(ErrorString, "Cannot use both --insert-pragma and --require-pragma") =/= nomatch).
```
michalmuskala, line 986: Is there a reason we're testing this specifically through stdio instead of the regular smoke test infrastructure we have?

src/erlfmt_cli.erl
```diff
@@ -18,6 +18,7 @@
 -record(config, {
     verbose = false :: boolean(),
     require_pragma = false :: boolean(),
+    insert_pragma = false :: boolean(),
```
michalmuskala, line 21: I wonder if it would make sense to also have a single `pragma = ignore :: ignore | require | insert` option here. We could avoid the extra conversion before calling `erlfmt:format_file`. Validation we're not using both option could be done in case we're setting an option when `Pragma =/= ignore`.

src/erlfmt.erl
```diff
@@ -68,8 +70,12 @@ format_file(FileName, {Pragma, Out}) ->
     try
         case file_read_nodes(FileName, Pragma) of
             {ok, Nodes, Warnings} ->
-                [$\n | Formatted] = format_nodes(Nodes),
-                verify_nodes(FileName, Nodes, Formatted),
+                NodesWithPragma = case Pragma of
+                    insert -> insert_pragma_nodes(Nodes);
+                    _ -> Nodes
+                end,
```
michalmuskala, line 76: What do you think about folding this check into `file_read_nodes`? We could add another clause here: https://github.com/WhatsApp/erlfmt/pull/63/files#diff-b4326b25dd3afce6a964f37ad5778582R207 with something like `false when Pragma =:= insert`

awalterschulze: I think then we would have to come up with a new name for file_read_nodes that makes it clear that it will also edit the nodes.  I cannot think of a good name, which is typically a warning sign, so I will keep this outside.  But if you come up with a name, please let me know.

src/erlfmt.erl
awalterschulze: I have updated this and added tests for rebar.config and .escript files.  Still working on other comments.

test/erlfmt_SUITE.erl
awalterschulze: I have moved `erlfmt:format_string` into erlfmt and removed the exports for `erlfmt:insert_pragma_nodes` and `erlfmt:format_nodes`

src/erlfmt_cli.erl
awalterschulze: I think this fixed in https://github.com/WhatsApp/erlfmt/pull/63/commits/7c2cfc138d505033f88ca55b157af3fb2aa80192

src/erlfmt_scan.erl
```diff
@@ -185,6 +191,8 @@ split_shebang(Tokens) ->
     {unicode:characters_to_list(Shebang), Rest}.
 
 %% TODO: make smarter
+stringify_tokens([]) ->
+    {"", undefined};
```
awalterschulze, line 195: This seems wrong, maybe it is time to make it smarter.
@michalmuskala do you have any ideas?

src/erlfmt_scan.erl
```diff
@@ -147,6 +147,10 @@ read_rest(#state{scan = _Scan, inner = IO, loc = _Loc, buffer = Buffer}) ->
         {error, Reason} -> {error, {0, file, Reason}}
     end.
 
+read_rest(undefined, Data) ->
```
awalterschulze, line 150: Still struggling to figure out how read_rest receives an `undefined` atom in the first place.

test/erlfmt_SUITE.erl
awalterschulze: Done https://github.com/WhatsApp/erlfmt/pull/63/commits/eade985ed41b32fb495fa30cf636007b54e7d2a0

test/erlfmt_SUITE.erl
awalterschulze: No reason specifically, it was just easy.  I will move most to snapshot testing, I assume that is what you mean with regular smoke test infrastructure.

awalterschulze: Hmmm, looking at the snapshot setup again, I think I picked the stdio version, because it is the only one that invokes the command with `os:cmd` and I want to test everything including the argument parsing.

PR #70, opened 2020-07-23
Fix require pragma for some edge cases

michalmuskala:
These include:
* non-unicode codepoints in the file
* single form with trailing comment

### DIFF
--- a/src/erlfmt_scan.erl
+++ b/src/erlfmt_scan.erl
@@ -131,8 +131,12 @@ continue(Scan, Inner0, Loc0, Buffer0) ->
     end.
 
 -spec read_rest(state()) -> {ok, string()} | {error, {erl_anno:location(), module(), term()}}.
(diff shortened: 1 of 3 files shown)


src/erlfmt_scan.erl
```diff
@@ -172,21 +178,12 @@ split_shebang(Tokens) ->
         end,
         Tokens
     ),
-    {Shebang, _Anno} = stringify_tokens(ShebangTokens),
+    Shebang = stringify_tokens(ShebangTokens),
     {unicode:characters_to_list(Shebang), Rest}.
 
 %% TODO: make smarter
```
awalterschulze, line 184: Wow can this really be simplified this way.
Is this the something smarter?  Do we still need the comment.

michalmuskala: I'm not really sure what I had in mind when I wrote that comment 😅
I'll remove it 👍

PR #73, opened 2020-07-23
Fix unicode handling for stdio and add more unicode tests

michalmuskala:
We weren't running stdio in unicode mode, which meant that reading from
it would sometimes misbehave. This was only visible when reading in `read_rest`
if we entered the loop on second iteration and onward, hence why the need for
the huge test fixture file.


test/erlfmt_SUITE_data/unicode.erl
```diff
@@ -0,0 +1,104 @@
...
+        1412, 1413, 1414, 1415, 1416, 1417, 1418, 1419, 1420, 1421, 1422, 1423, 1424, 1425,
+        1426, 1427, 1428, 1429, 1430, 1431, 1432, 1433, 1434, 1435, 1436, 1437, 1438, 1439,
+        1440, 1441, 1442, 1443, 1444, 1445, 1446, 1447, 1448, 1449, 1450, 1451, 1452, 1453,
+        1454, 1455, 1456, 1457, 1458, 1459, 1460, 1461, 1462, 1463, 1464, 1465, 1466, 1467,
+        1468, 1469, 1470, 1471, 1472, 1473, 1474, 1475, 1476, 1477, 1478, 1479, 1480, 1481,
+        1482, 1483, 1484, 1485, 1486, 1487, 1488, 1489, 1490, 1491, 1492, 1493, 1494, 1495,
+        1496, 1497, 1498, 1499, 1500>>.
+
+encode_decode_unicode(_Config) ->
+    "שלום עולם".
```
michalmuskala, line 104: I verified that the test fails correctly without the fix

awalterschulze:
Great job :D

PR #75, opened 2020-07-24
--insert-pragma inserts an @format comment into a file

### LINKED ISSUE #15: Add @format pragma handling

awalterschulze:
Replaces https://github.com/WhatsApp/erlfmt/pull/63
since both rebasing and merging lost too much information to recover.

Fixes #15

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -18,7 +18,7 @@
     main/1,
     init/1,
     format_file/2,
-    format_string/2, format_string/3,
+    format_string/3,
     format_range/3,
     read_nodes/1,
     read_nodes_string/2,
@@ -81,10 +81,6 @@ format_file(FileName, {Pragma, Out}) ->
     catch
         {error, Error} -> {error, Error}
     end.
-
--spec format_string(string(), integer()) -> string().
-format_string(String, PageWidth) ->
-    format_string(String, PageWidth, ignore).
 
 -spec format_string(string(), integer(), pragma()) -> string().
 format_string(String, PageWidth, Pragma) ->
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -18,13 +18,12 @@
     main/1,
     init/1,
     format_file/2,
-    format_string/2,
+    format_string/2, format_string/3,
     format_range/3,
     read_nodes/1,
     read_nodes_string/2,
     format_error/1,
     format_error_info/1,
-    contains_pragma_file/1,
     contains_pragma_string/2
 ]).
 
@@ -32,7 +31,7 @@
 
 -type error_info() :: {file:name_all(), erl_anno:location(), module(), Reason :: any()}.
 -type out() :: standard_out | {path, file:name_all()} | replace.
--type pragma() :: require | ignore.
+-type pragma() :: require | insert | ignore.
 -type config() :: {Pragma :: pragma(), Out :: out()}.
 
 -define(PAGE_WIDTH, 92).
@@ -66,8 +65,13 @@ format_file(FileName, {Pragma, Out}) ->
     try
         case file_read_nodes(FileName, Pragma) of
             {ok, Nodes, Warnings} ->
-                [$\n | Formatted] = format_nodes(Nodes, ?PAGE_WIDTH),
-                verify_nodes(FileName, Nodes, Formatted),
+                NodesWithPragma =
+                    case Pragma of
+                        insert -> insert_pragma_nodes(Nodes);
+                        _ -> Nodes
+                    end,
+                [$\n | Formatted] = format_nodes(NodesWithPragma, ?PAGE_WIDTH),
+                verify_nodes(FileName, NodesWithPragma, Formatted),
                 write_formatted(FileName, Formatted, Out),
                 {ok, Warnings};
             {skip, RawString} ->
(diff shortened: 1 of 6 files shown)


src/erlfmt_scan.erl
```diff
@@ -166,10 +168,14 @@ eof(undefined, Loc) ->
     {{eof, Loc}, undefined}.
 
 -spec last_node_string(state()) -> {unicode:chardata(), anno()}.
+last_node_string(#state{original = [{'#', _}, {'!', _} | _] = Tokens}) ->
+    split_shebang(Tokens);
```
michalmuskala, line 172: Why do we need this special case here?

src/erlfmt_scan.erl
```diff
@@ -166,10 +168,14 @@ eof(undefined, Loc) ->
...
 last_node_string(#state{original = [First | _] = Tokens}) ->
     String = stringify_tokens(Tokens),
     Location = erl_scan:location(First),
-    {String, token_anno([{text, String}, {location, Location}])}.
+    {String, token_anno([{text, String}, {location, Location}])};
+last_node_string(#state{original = []}) ->
+    {"", undefined}.
```
michalmuskala, line 178: When can this option happen?

src/erlfmt_scan.erl
awalterschulze: You are right, I needed it before, but in this new version, I don't need it anymore.  Good catch, I removed it.

test/erlfmt_SUITE.erl
```diff
@@ -1053,18 +1083,18 @@ contains_pragma(Config) when is_list(Config) ->
     ?assertEqual(
         true,
         erlfmt:contains_pragma_string(
```
michalmuskala, line 1085: Looking at those tests, I'm not really sure about their value. In particular we don't use `erlfmt:contains_pragma_string` in regular code, so this seems like it's testing a function exported only for tests

src/erlfmt_scan.erl
awalterschulze: It is needed for these cases
```erlang

erlfmt:contains_pragma_string(
    "file.escript",
    "#! /usr/bin/env escript\n"
    "\n"
    "%% @format\n"
    "\n"
    "main(_) -> ok.\n"
)
erlfmt:contains_pragma_string(
    "file.escript",
    "#! /usr/bin/env escript\n"
    "\n"
    "main(_) -> ok.\n"
)
```

src/erlfmt.erl
```diff
@@ -80,37 +84,73 @@ format_file(FileName, {Pragma, Out}) ->
 
 -spec format_string(string(), integer()) -> string().
 format_string(String, PageWidth) ->
+    format_string(String, PageWidth, ignore).
```
michalmuskala, line 87: Do we need to keep the `format_string/2` function around? I'd be fine only having the `/3` variant. For formatting file we don't have a version without options.

test/erlfmt_SUITE.erl
awalterschulze: contains_pragma_string is calling `read_nodes(erlfmt_scan:string_node(String), FileName, require)` much like the regular code is doing.  At least this is the idea.

src/erlfmt.erl
awalterschulze: removed :)

src/erlfmt_scan.erl
michalmuskala: Good catch!

I think this is because we're not setting `original` correctly when reading the shebang in [here](https://github.com/WhatsApp/erlfmt/blob/b95bc12e037001da4bc447f742c0f06c4dfebf8a/src/erlfmt_scan.erl#L92). I think a nicer fix would be to modify `split_shebang` to return the tokens, setting `original` in there and we wouldn't need any special-casing in here.

src/erlfmt.erl
```diff
@@ -78,39 +82,71 @@ format_file(FileName, {Pragma, Out}) ->
...
+        {error, Error} -> {error, Error}
+    end.
 
-read_pragma_nodes({ok, Tokens, Comments, Cont}, FileName, _Acc, Warnings0) ->
-    {Node, _Warnings} = parse_nodes(Tokens, Comments, FileName, Cont, Warnings0),
-    contains_pragma_node(Node);
-read_pragma_nodes(_, _, _, _) ->
-    false.
+is_shebang_node({raw_string, _, "#!" ++ _}) -> true;
+is_shebang_node(_) -> false.
```
michalmuskala, line 106: We're converting a special `shebang` node into raw string in here https://github.com/WhatsApp/erlfmt/blob/1b3a001e1f64616a4b484ef4a15bbf681e81062d/src/erlfmt.erl#L255-L256. Do you think it would make sense to keep the special node around longer to simplify this function or even remove it and pattern match on the `shebang` node directly? We would need to change `format_node` to handle the new `shebang` in the same way that it's handling `raw_string` right now

awalterschulze: That sounds great.  How about I tackle that in a follow up commit?

src/erlfmt_scan.erl
awalterschulze: Yes, this is amazing.  That for the great idea :D

src/erlfmt.erl
awalterschulze: On second thought, I saw it was easier than expected.  So I have now added it in this commit.

PR #79, opened 2020-07-24
Correctly handle .script files

michalmuskala:
They can have multiple expressions between each dot, unlike .config files.
Add a special top-level exprs node to handle that situation.


test/erlfmt_SUITE_data/rebar.config.script
```diff
@@ -0,0 +1,14 @@
+%% Example from https://github.com/benoitc/unicode_util_compat/blob/c39aed13801374aa577b5d75a93384276f49e8aa/rebar.config.script
```
awalterschulze, line 1: Oh wow, this probably has implications for insert_pragma, contains_pragma, but I can follow up in a different commit.

michalmuskala: I think it should be fine. There's still one node and we insert the comment before it. Since we support any node, not only an attribute as the first one, I think it should "just work".

awalterschulze: Ah yes, you made me fix that before.  Good point :)

PR #82, opened 2020-07-30
erlfmt_format: single line clauses with comments, stay single line clauses

### LINKED ISSUE #76: Investigate unnecessary multiline layout for clauses with comments
```erlang
receive
    _ -> ok % foo
end
```
becomes
```erlang
receive
    % foo
    _ ->
        ok
 end
```
while it should become:
```erlang
receive
    % foo
    _ -> ok
 end
```

awalterschulze:
Helps with #76 unnecessary multiline layout for clauses with comments

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -235,7 +235,7 @@ surround(Left, LeftSpace, Doc, RightSpace, Right) ->
 surround_block(Left, Doc, Right) ->
     concat(
         force_breaks(),
-        (group(line(concat(Left, nest(concat(line(), Doc), ?INDENT)), Right)))
+        group(line(concat(Left, nest(concat(line(), Doc), ?INDENT)), Right))
     ).
 
 string_to_algebra(Text) ->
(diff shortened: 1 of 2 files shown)


src/erlfmt_format.erl
```diff
@@ -528,7 +528,24 @@ fun_to_algebra({type, Meta, Args, Result}) ->
...
+        true -> clauses_to_algebra_([Clause]);
+        false ->
+            %% This is a special case for a single clause,
+            %% with none of its own breaks.
+            %% We want to keep it in a single line,
+            %% even if it has comments that would usually force breaks using:
+            %% `group(Doc)`
+            Meta = element(2, Clause),
+            Doc = do_expr_to_algebra(Clause),
+            combine_comments(Meta, maybe_wrap_in_parens(Meta, group(Doc)))
```
michalmuskala, line 543: Will this work correctly for something like following? 
```erl
receive
    1 -> one; % one
    2 -> two % two
end.
```

This also shouldn't break, I think it should be formatted as:
```erl
receive
    % one
    1 -> one;
    % two
    2 -> two
end.
```

awalterschulze: Both are formatted to
```erlang
receive
    % one
    1 ->
        one;
    % two
    2 ->
        two
end.
```

awalterschulze: The intention was only to handle the single clause case, I could extend it to handle more than one clause, if you think that would be better.

michalmuskala: Yes, I think we should. This wasn't clear in the issue I created, but I intended for it to work like that for multiple clauses as well. I think we should merge this as is for now, keep the issue open, and solve the multi-clause case next.

awalterschulze: Sounds good, I updated the description not to say "fixes".

PR #83, opened 2020-07-31
erlfmt_format: preserve empty lines between values in containers

### LINKED ISSUE #72: Preserve empty lines in export lists
```erlang
-export([
  % comment
  a/2,

  % another group
  b/1
]).
```

awalterschulze:
Fixes #72

This preserves empty lines between values in containers, like exports and records.

This required some refactoring, since the simple fold_doc now needed to not only be aware of the break, but also needed to have the original value, to be able to check for empty lines to preserve.

Really struggled to make this refactoring more readable and I guess it is debatable, whether I achieved that, so looking forward to ideas in the review.

Also needed to preserve the end line for better empty line detection.
Should probably do the same for inner end line, so maybe we can think of a better function name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,24 +214,18 @@ the collection will be always printed "expanded", for example:
 ]
 ```
 
-will be preserved, even though it could fit on a single line. The fact that
-this is controlled by a single newline, allows you to easily convert between
-those layouts, for example, merely deleting the first newline from the above
-sequence to have:
+will be preserved, even though it could fit on a single line.
+This is controlled by whether the user has included a newline in the original version.
+For example, merely deleting the newlines from the above sequence:
 
 ```erl unformatted foobar1
-[    Foo,
-    Bar
-]
+[    Foo, Bar]
 ```
 
 and re-running the formatter, will produce:
 
 ```erl formatted foobar1
-[
-    Foo,
-    Bar
-]
+[Foo, Bar]
 ```
 
 Similarly, adding the single newline back:
--- a/README.md
+++ b/README.md
@@ -228,7 +228,10 @@ sequence to have:
 and re-running the formatter, will produce:
 
 ```erl formatted foobar1
-[Foo, Bar]
+[
+    Foo,
+    Bar
+]
 ```
 
 Similarly, adding the single newline back:
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -349,11 +349,17 @@ container_common(_Meta, [], Left, Right, _, _) ->
 container_common(Meta, Values, Left, Right, Break, LastFits0) ->
     SplitValues = init_last_and_comments(Values),
     LastFits = last_fits(LastFits0, SplitValues),
-    HasLineBreak = has_line_break(Meta, SplitValues),
-    BreakFun = break_fun(Break, HasLineBreak),
-    SurroundFun = surround_fun(HasLineBreak, Break, LastFits),
+    {BreakFun, SurroundFun} = surround_and_break_fun(Meta, SplitValues, Break, LastFits),
     Doc = container_inner_values(SplitValues, BreakFun, LastFits),
     SurroundFun(Left, Doc, Right).
+
+surround_and_break_fun(Meta, SplitValues, Break, LastFits) ->
+    HasMetaLineBreak = has_meta_line_break(Meta, SplitValues),
+    HasAnyLineBreak = has_line_break(SplitValues),
+    {
+        break_fun(Break, HasMetaLineBreak),
+        surround_fun(HasMetaLineBreak, HasAnyLineBreak, Break, LastFits)
+    }.
 
 -record(split_values, {
     init_values :: [erlfmt_parse:abstract_form()],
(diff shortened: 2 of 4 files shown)


src/erlfmt_format.erl
```diff
@@ -421,6 +519,8 @@ fa_groups([{op, Meta, '/', {atom, _, HeadFunction}, _} = Head0 | Tail]) ->
             [Head0 | fa_groups(Rest)];
         {Group, Rest} ->
             Head = erlfmt_scan:delete_annos([pre_comments, post_comments], Head0),
+            EndLoc = erlfmt_scan:get_end_location(lists:last(Group)),
+            Meta = erlfmt_scan:set_end_location(EndLoc, Meta0),
```
michalmuskala, line 523: In the parser we have a `?range_anno` macro that does something similar. What do you think about having a similar function in `erlfmt_scan`? Something like:
```erl
range_anno(First, Last) when is_tuple(First), is_tuple(Last) ->
    (element(2, First))#{end_location := get_end_location(Last)}.
```

We could call it here as:
```erl
Meta = erlfmt_scan:range_anno(Head, lists:last(Group))
```

awalterschulze: That sounds great :)

awalterschulze: We should probably also set the `inner_end_location`, right?

michalmuskala:
Looking at all the complexity trailing comments introduce, I started thinking if maybe we should change how we represent them in the AST. My proposal would be to use the `post_comment` anno of the last element if the container has elements. If the container is empty with only the comment inside, we would still have the standalone comment node. What do you think? I believe it should make a lot of code simpler, we have to special-case those standalone comments a lot and effectively apply some manual processing that would be done for `post_comment` of last node anyway.

src/erlfmt_format.erl
michalmuskala: Hmm... potentially. Do we really need it? It would need some `case`ing to only add it if it was there in `Last` in the first place.

awalterschulze: I added range_anno, but I needed to pass in Meta, otherwise, we lose the comments.

awalterschulze:
> Looking at all the complexity trailing comments introduce, I started thinking if maybe we should change how we represent them in the AST. My proposal would be to use the `post_comment` anno of the last element if the container has elements. If the container is empty with only the comment inside, we would still have the standalone comment node. What do you think? I believe it should make a lot of code simpler, we have to special-case those standalone comments a lot and effectively apply some manual processing that would be done for `post_comment` of last node anyway.

I guess I have always thought the pre_comments were too attached. I would expect to have more standalone comments in the ast.  Basically if there is an empty line between this and the next element, then the comment is intuitively not attached to the next element. 
So that would be one way of refactoring, is to embrace the detached comments.

But to be pragmatic we could also embrace the aggressive attachment of comments and consistently say that all comments are attached.  This would attach these trailing comments as post comments.

Which ever way we go, I think we can do this in a follow up pull request, and possibly pair on it together.

michalmuskala:
The problem of standalone comments is that they don't behave syntactically like other elements, for example inside lists they don't have `,` in-between them. Attaching comments to elements (what your code effectively does in here) bypasses this issue.

awalterschulze:
Yes so I am happy to go with the pragmatic route.

src/erlfmt_format.erl
```diff
@@ -344,57 +344,155 @@ flex_container(Meta, Values, Left, Right) ->
...
-            case WasLastFits of
-                true -> next_break_fits(Wrapped, disabled);
-                false -> Wrapped
-            end
-    end.
+container_common(Meta, Values, Left, Right, Break, LastFits0) ->
+    SplitValues = init_last_and_comments(Values),
+    LastFits = last_fits(LastFits0, SplitValues),
+    HasLineBreak = has_line_break(Meta, SplitValues),
+    BreakFun = break_fun(Break, HasLineBreak),
```
michalmuskala, line 353: What do you think about having something like:
```erl
{BreakFun, SurroundFun} = break_and_surround_fun(Break, Meta, SplitValues)
```
It seems that they are both computed in a very similar way, but now it's split in two places.

This would allow avoiding passing `HasLineBreak` and `Break` arguments to `surround_container_values` in favour of single `SurroundFun` and moving `has_line_break` call inside of `break_and_surround_fun`.

michalmuskala:
What happens if the way the container was initially laid out didn't include the "force breaks" space? In particular, if it was written down in a "comma-first" style like below:

```erl
[ 1
, 2

, 3
]
```

Do we collapse it into a single line `[1, 2, 3]` or do we expand it into:
```erl
[
    1,
    2,

    3
]
```

Either choice is fine by me, I think we should have a test that explores this, though.

src/erlfmt_format.erl
awalterschulze: I can add SurroundFun, I agree that would look much better, but would prefer not to create a function that does two things `break_and_surround_fun`.  But I will also see what it looks like in practice.

michalmuskala: Lumping them together for me makes sense since they use exactly the same conditions. I agree the name is not that nice.

awalterschulze:
> a test that explores this, though.

Great catch :)
This seems to be the current passing test.

```erlang
?assertFormat(
        "[ 1\n"
        ", 2\n"
        "\n"
        ", 3\n"
        "]\n",
        "[1, 2,\n"
        "\n"
        "3]\n"
    ).
```

I also agree, that either case would be great, except the one that is happening now.

src/erlfmt_format.erl
awalterschulze: I pushed the commit that doesn't lump them together, but also tried to lump them together, here is what the diff would be.

```
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -349,12 +349,14 @@ container_common(_Meta, [], Left, Right, _, _) ->
 container_common(Meta, Values, Left, Right, Break, LastFits0) ->
     SplitValues = init_last_and_comments(Values),
     LastFits = last_fits(LastFits0, SplitValues),
-    HasLineBreak = has_line_break(Meta, SplitValues),
-    BreakFun = break_fun(Break, HasLineBreak),
-    SurroundFun = surround_fun(HasLineBreak, Break, LastFits),
+    {BreakFun, SurroundFun} = surround_and_break_fun(Meta, SplitValues, Break, LastFits),
     Doc = container_inner_values(SplitValues, BreakFun, LastFits),
     SurroundFun(Left, Doc, Right).
 
+surround_and_break_fun(Meta, SplitValues, Break, LastFits) ->
+    HasLineBreak = has_line_break(Meta, SplitValues),
+    {break_fun(Break, HasLineBreak), surround_fun(HasLineBreak, Break, LastFits)}.
+
```

awalterschulze: Needed to another type of HasBreak and now surround_and_break_fun started having more of a benefit, so I added it.

awalterschulze:
Have to say it is very tough to get this and other containers like tuples, to place nice together

> What happens if the way the container was initially laid out didn't include the "force breaks" space between `[` and first element? In particular, if it was written down in a "comma-first" style like below:
> 
> ```erlang
> [ 1
> , 2
> 
> , 3
> ]
> ```
> 
> Do we collapse it into a single line `[1, 2, 3]` or do we expand it into:
> 
> ```erlang
> [
>     1,
>     2,
> 
>     3
> ]
> ```
> 
> Either choice is fine by me, I think we should have a test that explores this, though.

This is now fixed, it was quite tricky :)

README.md
```diff
@@ -228,7 +228,10 @@ sequence to have:
 and re-running the formatter, will produce:
 
 ```erl formatted foobar1
-[Foo, Bar]
+[
+    Foo,
+    Bar
+]
```
michalmuskala, line 234: Hmm.... We should probably re-write the guide because it no longer works as advertised. It's not enough to just remove the newline between `[` and first element.

This is so much complexity in here 😢

michalmuskala:
> Have to say it is very tough to get this and other containers like tuples, to place nice together

Yeah, there's a lot of complexity in the rules for different containers. We probably need to rethink the general approach of keeping everything together after merging this code.

README.md
awalterschulze: Good catch.
I fixed the guide.

I think it makes sense to not only take a hint from the first newline, but rather from any newline in the container.

PR #84, opened 2020-07-31
erlfmt_format: dont group pipes so they all break together

### LINKED ISSUE #53: Typespec variants not kept on separate lines
(erlfmt @ e81bcb4)

When splitting the variants of an argument to a function typespec over several lines, it seems that erlfmt may leave more than one case in the last line:

```diff
--spec encode(#user_creation_token{} | #user_authentication_token{} | #user_password_reset_token{} | #user_email_verification_token{}) -> binary().
+-spec encode(
+    #user_creation_token{} |
+    #user_authentication_token{} |
+    #user_password_reset_token{} | #user_email_verification_token{}
+) ->
+    binary().
```

awalterschulze:
Fixes #53

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -311,10 +311,18 @@ binary_op_to_algebra(Op, Meta, Left, Right, Indent) ->
     IsNextBreakFits = is_next_break_fits_op(Op, Right) andalso not HasBreak,
 
     Doc =
-        with_next_break_fits(IsNextBreakFits, RightD, fun (R) ->
-            RightOpD = nest(break(concat(<<" ">>, OpD), group(R)), Indent, break),
-            concat(group(LeftD), group(concat(maybe_force_breaks(HasBreak), RightOpD)))
-        end),
+        case Op of
+            '|' ->
+                 with_next_break_fits(IsNextBreakFits, RightD, fun (R) ->
+                    RightOpD = nest(break(concat(<<" ">>, OpD), R), Indent, break),
+                    concat(LeftD, concat(maybe_force_breaks(HasBreak), RightOpD))
+                end);
+            _ ->
+                with_next_break_fits(IsNextBreakFits, RightD, fun (R) ->
+                    RightOpD = nest(break(concat(<<" ">>, OpD), group(R)), Indent, break),
+                    concat(group(LeftD), group(concat(maybe_force_breaks(HasBreak), RightOpD)))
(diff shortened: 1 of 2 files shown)


src/erlfmt_format.erl
```diff
@@ -311,10 +311,18 @@ binary_op_to_algebra(Op, Meta, Left, Right, Indent) ->
...
     Doc =
-        with_next_break_fits(IsNextBreakFits, RightD, fun (R) ->
-            RightOpD = nest(break(concat(<<" ">>, OpD), group(R)), Indent, break),
-            concat(group(LeftD), group(concat(maybe_force_breaks(HasBreak), RightOpD)))
-        end),
+        case Op of
+            '|' ->
+                 with_next_break_fits(IsNextBreakFits, RightD, fun (R) ->
+                    RightOpD = nest(break(concat(<<" ">>, OpD), R), Indent, break),
+                    concat(LeftD, concat(maybe_force_breaks(HasBreak), RightOpD))
```
michalmuskala, line 318: I'm not sure this would behave correctly with mixed operators. For example something like `1 + 2 | 3 + 4`

awalterschulze: This seems to be a parse error `"1 + 2 | 3 + 4\n"` and `"Foo = 1 + 2 | 3 + 4\n"`
I didn't know you could write that, do you have a parsable example?

michalmuskala: I'm sorry. This should be a complete example that parses fine today:
```erl
-type foo() :: 1 + 2 | 3 + 4.
```

awalterschulze: These both work
```erlang
?assertSame("-type foo() :: 1 + 2 | 3 + 4.\n"),
    ?assertSame(
        "-type foo() ::\n"
        "    1 + 2 |\n"
        "    3 + 4.\n"
    ).
```

awalterschulze: I could add this one as a test
```erlang
?assertSame(
    "-type foo() ::\n"
    "    1 + 2 |\n"
    "    3 + 4.\n"
).
```
But maybe you have a different test in mind?

michalmuskala: Looks good. It seems the thing I was worried about works correctly. Great job!

awalterschulze: I got Lucky :)

PR #90, opened 2020-08-05
Fix multiline type unions formatting

### LINKED ISSUE #89: Alternatives within typespec argument lists are always split over multiple lines
erlfmt 0.5.0

(This might be related to #53.)

When formatting with 0.5.0 (after upgrading from 0.4.2), I saw a surprising number of changed files, and it turns out that the changes are typespecs' arguments list where alternatives now always are split over several lines:

```diff
 -spec use_credit(
     store:id(),
     decimal:decimal(),
-    binary() | null,
-    credit | {refund | dispute, binary()},
-    binary() | null
+    binary() |
+    null,
+    credit |
+    {refund | dispute, binary()},
+    binary() |
+    null
 ) -> ok.
```

Is this intentional? I would expect e.g. `binary() | null` to be left on a single line here like in 0.4.2, so this code would remain like this:

```erlang
 -spec use_credit(
     store:id(),
     decimal:decimal(),
     binary() | null,
     credit | {refund | dispute, binary()},
     binary() | null
 ) -> ok.
```

In #53, what happened was that a sequence of alternatives (`|`) needed to be split over multiple lines, and then _all_ alternatives should be on separate lines, but in this case, it's the arguments (`,`) that need to be split into separate lines, not the alternatives _within_ an argument.

Hope this made sense. 😄

michalmuskala:
Separate formatting of type unions from other operators - their format
seems distinct enough to warrant this and the code looks simpler.

Closes #89

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -294,11 +294,9 @@ binary_op_to_algebra('/', _Meta, {atom, _, _} = Left, {integer, _, _} = Right) -
     %% special-case the foo/N syntax in attributes, division with a literal atom
     %% doesn't make sense in normal code, so it's safe to apply it everywhere
     concat(expr_to_algebra(Left), <<"/">>, expr_to_algebra(Right));
-binary_op_to_algebra('|' = Op, Meta0, Left, Right) ->
-    %% | does not increase indentation
-    %% don't print parens and comments twice - expr_to_algebra took care of it already
-    Meta = erlfmt_scan:delete_annos([parens, pre_comments, post_comments], Meta0),
-    binary_op_to_algebra(Op, Meta, Left, Right, 0);
+binary_op_to_algebra('|', Meta, Left, Right) ->
+    %% | does not increase indentation and breaks as a whole
+    union_to_algebra(Meta, Left, Right);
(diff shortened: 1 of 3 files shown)


src/erlfmt_format.erl
```diff
@@ -312,21 +310,23 @@ binary_op_to_algebra(Op, Meta, Left, Right, Indent) ->
...
-        end,
-
+        with_next_break_fits(IsNextBreakFits, RightD, fun (R) ->
+            RightOpD = nest(break(concat(<<" ">>, OpD), group(R)), Indent, break),
+            concat(group(LeftD), group(concat(maybe_force_breaks(HasBreak), RightOpD)))
+        end),
     combine_comments(Meta, maybe_wrap_in_parens(Meta, Doc)).
 
+union_to_algebra(Meta, Left, Right) ->
+    Unions = [Left | collect_unions(Right)],
```
awalterschulze, line 320: I guess we are relying a bit on how the parser disambiguates by not calling `collect_unions(Left)`

michalmuskala: Yeah, we rely on the fact that it's right-associative and recurse only on the right side. If there is a nested union on the left as well, it had to be wrapped in parens to even parse and probably should be treated as a separate expression.

PR #91, opened 2020-08-07
Add references to Elvis/Linter rules.

elbrujohalcon:
Using `rebar3_lint` / `elvis`, you can automate most of the "manual" work required for linting.
I think it's a good idea to guide folks that reach this page from the broader Erlang ecosystem to a tool that can help them with that, just like `erlfmt` can help them with formatting.
I would also like to recommend you to take a look at the [Erlang Guidelines](https://github.com/inaka/erlang_guidelines). Maybe you can contribute some updates based on what's here and we can unify the list, to provide a single comprehensive guide for newcommers.

### DIFF
--- a/StyleGuide.md
+++ b/StyleGuide.md
@@ -10,6 +10,7 @@ This style guide represents guidelines for writing good Erlang code according
 to the authors of erlfmt. It consists of two sections:
   * Linting - with guidelines about code content that are considered good
     practice and need to be applied manually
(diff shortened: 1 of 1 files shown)


michalmuskala:
Thank you! ♥️

michalmuskala:
I like the idea of trying to unify this list with the guidelines. I'll take a look and see what we could do in there.

Do you think there are any major differences between the two?

elbrujohalcon:
Not sure. You have some really good ones that we don't have. That's for sure.
And to be fair, some of the guidelines in inaka's repo might be outdated.

PR #123, opened 2020-09-01
add --check option

awalterschulze:
Test Plan:
```sh
$ _build/release/bin/erlfmt --check "{src,include,test}/*.{hrl,erl}" "rebar.config"
Checking formatting...
[warn] src/erlfmt_cli.erl
[warn] src/erlfmt.erl
[warn] src/erlfmt_format.erl
[warn] test/erlfmt_format_SUITE.erl
[warn] src/erlfmt_parse.erl
[warn] Code style issues found in the above file(s). Forgot to run erlfmt?
walterschul-mbp:erlfmt walterschulze$ echo $?
1
```

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -29,7 +29,7 @@
 -export_type([error_info/0, out/0, config/0, pragma/0]).
 
 -type error_info() :: {file:name_all(), erl_anno:location(), module(), Reason :: any()}.
--type out() :: standard_out | {path, file:name_all()} | replace.
+-type out() :: standard_out | {path, file:name_all()} | replace | check.
 -type pragma() :: require | insert | ignore.
 -type config() :: [{pragma, pragma()} | {width, pos_integer()}].
 
@@ -59,7 +59,11 @@ init(State) ->
 
 %% API entry point
 -spec format_file(file:name_all() | stdin, out(), config()) ->
-    {ok, [error_info()]} | skip | {error, error_info()}.
+    {ok, [error_info()]} |
+    {check_failed, string(), string(), [error_info()]} |
+    {[error_info()]} |
+    skip |
+    {error, error_info()}.
 format_file(FileName, Out, Options) ->
     Width = proplists:get_value(width, Options, ?DEFAULT_WIDTH),
     Pragma = proplists:get_value(pragma, Options, ignore),
@@ -73,8 +77,10 @@ format_file(FileName, Out, Options) ->
                     end,
                 [$\n | Formatted] = format_nodes(NodesWithPragma, Width),
                 verify_nodes(FileName, NodesWithPragma, Formatted),
-                write_formatted(FileName, Formatted, Out),
-                {ok, Warnings};
+                case write_formatted(FileName, Formatted, Out) of
+                    {check_failed, OriginalStr, FormattedStr} -> {check_failed, OriginalStr, FormattedStr, Warnings};
+                    ok -> {ok, Warnings}
+                end;
             {skip, RawString} ->
                 write_formatted(FileName, RawString, Out),
                 skip
@@ -423,6 +429,16 @@ try_location(_, Node) when is_tuple(Node) -> erlfmt_scan:get_anno(location, Node
 try_location(_, [Node | _]) when is_tuple(Node) -> erlfmt_scan:get_anno(location, Node);
 try_location(_, _) -> 0.
 
+write_formatted(FileName, Formatted, check) ->
+    Original = read_file(FileName, fun(File) ->
+        erlfmt_scan:read_all(File)
+    end),
+    FormattedStr = unicode:characters_to_list(Formatted),
+    OriginalStr = unicode:characters_to_list(Original),
+    case OriginalStr =:= FormattedStr of
+        true -> ok;
+        false -> {check_failed, OriginalStr, FormattedStr}
+    end;
 write_formatted(_FileName, Formatted, standard_out) ->
     io:put_chars(Formatted);
 write_formatted(FileName, Formatted, Out) ->
--- a/src/erlfmt_cli.erl
+++ b/src/erlfmt_cli.erl
@@ -30,6 +30,11 @@ opts() ->
         {write, $w, "write", undefined, "modify formatted files in place"},
         {out, $o, "out", binary, "output directory"},
         {verbose, undefined, "verbose", undefined, "include debug output"},
+        {check, $c, "check", undefined,
+            "Check if your files are formatted."
+            "Get exit code 1, if some files are not formatted."
+            "stdin is not supported and --write is not supported."
+        },
         {print_width, undefined, "print-width", integer,
             "The line length that formatter would wrap on"},
         {require_pragma, undefined, "require-pragma", undefined,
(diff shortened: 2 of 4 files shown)


src/erlfmt.erl
```diff
@@ -59,7 +59,11 @@ init(State) ->
 
 %% API entry point
 -spec format_file(file:name_all() | stdin, out(), config()) ->
-    {ok, [error_info()]} | skip | {error, error_info()}.
+    {ok, [error_info()]} |
+    {check_failed, string(), string(), [error_info()]} |
+    {[error_info()]} |
```
michalmuskala, line 64: In what case do we return just the bare warnings?

src/erlfmt.erl
```diff
@@ -423,6 +429,16 @@ try_location(_, Node) when is_tuple(Node) -> erlfmt_scan:get_anno(location, Node
 try_location(_, [Node | _]) when is_tuple(Node) -> erlfmt_scan:get_anno(location, Node);
 try_location(_, _) -> 0.
 
+write_formatted(FileName, Formatted, check) ->
+    Original = read_file(FileName, fun(File) ->
+        erlfmt_scan:read_all(File)
+    end),
```
michalmuskala, line 435: We can use `file:read_file/1` here

```suggestion
    {ok, Original} = file:read_file(FileName),
```

src/erlfmt.erl
```diff
@@ -423,6 +429,16 @@ try_location(_, Node) when is_tuple(Node) -> erlfmt_scan:get_anno(location, Node
...
+    FormattedStr = unicode:characters_to_list(Formatted),
+    OriginalStr = unicode:characters_to_list(Original),
+    case OriginalStr =:= FormattedStr of
```
michalmuskala, line 438: With `file:read_file` we get a single binary back, we don't need to convert it, if we also convert formatted to a binary (this should also be slightly faster).
```suggestion
    FormattedStr = unicode:characters_to_binary(Formatted),
    case Original =:= FormattedStr of
```

test/erlfmt_SUITE.erl
```diff
@@ -1011,32 +1011,24 @@ snapshot_insert_pragma_with(Config) when is_list(Config) ->
...
-    ),
-    {ok, FormattedFormatted} =
-        file:read_file(filename:join([PrivDir, Module ++ ".formatted"])),
-    ?assertEqual(Expected, FormattedFormatted).
+    case erlfmt:format_file(filename:join([DataDir, Module]), check, []) of
+        {ok, _} -> ct:fail("expected ~p to require some formatting", [Module]);
+        skip -> ok;
+        {check_failed, _, Formatted, _} -> ?assertEqual(binary_to_list(Expected), Formatted);
+        Other -> ct:fail("unexpected: ~p~n", [Other])
+    end.
```
michalmuskala, line 1031: I believe this skips one check that was there originally - is the formatting idempotent

src/erlfmt_cli.erl
```diff
@@ -56,9 +61,16 @@ do(Opts, Name) ->
...
+                _ -> ok
+            end,
             case parallel(fun(File) -> format_file(File, Config) end, Files) of
-                true -> erlang:halt(4);
-                false -> ok
+                ok -> ok;
+                warn ->
+                    io:format(standard_error, "[warn] Code style issues found in the above file(s). Forgot to run erlfmt?~n", []),
+                    erlang:halt(1);
+                error -> erlang:halt(4)
```
michalmuskala, line 73: What happens if we get both errors and check failures? Does the output look reasonable?

michalmuskala (commented):
I think this looks good overall! Just couple minor questions

src/erlfmt.erl
awalterschulze: I started with that, but it doesn't work for unicode

test/erlfmt_SUITE.erl
awalterschulze: This is what `snapshot_same` is doing in the first line of the function.

src/erlfmt_cli.erl
awalterschulze: The error trumps the warning

src/erlfmt.erl
awalterschulze: Oh that is a typo good catch

src/erlfmt.erl
awalterschulze: See the above comment.

src/erlfmt_cli.erl
awalterschulze: So it would not print out the final warn and only print out the final error.  I think this is acceptable, what do you think?

src/erlfmt.erl
awalterschulze: But I will try it out again, just to double check

awalterschulze: You were right, again :)
I will push a fix.

src/erlfmt.erl
awalterschulze: Fixed, thanks again, great catch.

test/erlfmt_SUITE.erl
michalmuskala: Ah, I missed that. Makes sense

src/erlfmt_cli.erl
michalmuskala: Yeah, I think that's fine

michalmuskala (approved):
Looks good!

src/erlfmt.erl
```diff
@@ -59,7 +59,10 @@ init(State) ->
 
 %% API entry point
 -spec format_file(file:name_all() | stdin, out(), config()) ->
-    {ok, [error_info()]} | skip | {error, error_info()}.
+    {ok, [error_info()]} |
+    {check_failed, binary(), binary(), [error_info()]} |
```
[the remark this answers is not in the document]
michalmuskala: Yeah, I was thinking about this one too. It seemed to be used in tests, but perhaps that's not enough to justify such API

/cc @awalterschulze

src/erlfmt.erl
```diff
@@ -423,6 +428,13 @@ try_location(_, Node) when is_tuple(Node) -> erlfmt_scan:get_anno(location, Node
 try_location(_, [Node | _]) when is_tuple(Node) -> erlfmt_scan:get_anno(location, Node);
 try_location(_, _) -> 0.
 
+write_formatted(FileName, Formatted, check) ->
+    {ok, OriginalBin} = file:read_file(FileName),
+    FormattedBin = unicode:characters_to_binary(Formatted),
+    case OriginalBin =:= FormattedBin of
```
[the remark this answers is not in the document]
awalterschulze: Ha ha, Great idea, thank you for pointing it out, I like it :)

awalterschulze:
I can always make a follow up commit, not too late :)

src/erlfmt.erl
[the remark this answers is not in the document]
awalterschulze: This is very useful for testing and I was hoping this can be used to offload to a diffing function in future.
Then someone can pass a --diff flag and we only need to change erlfmt_cli.erl
But maybe this is not needed.

Another option is maybe to rather just read the file and pass it to format_string and then check if there is a diff and bypass using format_file all together

What do you think?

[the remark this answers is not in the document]
elbrujohalcon: If I were making decisions here, I would tell you that we can implement that when the need arises. I'm not a big fan of _futurology_ myself. But I don't have all the context that you have.
On my side, I can work with any API you like, but I do prefer the simpler ones ;)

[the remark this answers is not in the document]
awalterschulze: I have made adjustments to rather use format_string, which also simplifies the API
And started using more pattern matching instead of equals
https://github.com/WhatsApp/erlfmt/pull/133/

PR #124, opened 2020-09-02
another option to discuss for comments

awalterschulze:


ilya-klyuchnikov:
Let me add one more option here:
There is still 48, and comments are still aligned! - We just use an extra line)

```erlang
f() ->
    g(
        #{
            A => a,                           % comment
            B => bb,                          % another comment
                                              % that comment
            CThisLongerField => with_this_longer_function(), 
            NotTooLong => d                   % this comment
        },
        H,
    ).
```

ilya-klyuchnikov:
Ah, - there is another option here - to put this hanging comment after the line, - Then the "mental structure" is preserved: comments are always after the stuff:

```
f() ->
    g(
        #{
            A => a,                           % comment
            B => bb,                          % another comment
            CThisLongerField => with_this_longer_function(), 
                                              % that comment for with_this_longer_function
            NotTooLong => d                   % this comment
        },
        H,
    ).
```

awalterschulze:
Thank you @ilya-klyuchnikov I have now also added your options to the document :)

david-kubecka:
Not sure if this PR (vs. an issue) is the correct place to discuss but I was redirected here from https://github.com/WhatsApp/erlfmt/issues/146 :-)

It seems that regarding inline comments you don't consider the option outlined [here](https://github.com/WhatsApp/erlfmt/blob/master/doc/FormattingDecisionComments.md#single-comments-on-the-same-line). Why? :-) It seems to me to be highly consistent, easy to implement and potentially solving other problems, e.g. most of those mentioned [here](https://github.com/WhatsApp/erlfmt/issues/134).

awalterschulze:
> Not sure if this PR (vs. an issue) is the correct place to discuss but I was redirected here from #146 :-)
> 
> It seems that regarding inline comments you don't consider the option outlined [here](https://github.com/WhatsApp/erlfmt/blob/master/doc/FormattingDecisionComments.md#single-comments-on-the-same-line). Why? :-) It seems to me to be highly consistent, easy to implement and potentially solving other problems, e.g. most of those mentioned [here](https://github.com/WhatsApp/erlfmt/issues/134).

@david-kubecka yes we still need to make it clear which option we are picking.
I have add a TODO to hopefully make this more clear.

awalterschulze:
I think we should merge this as the current state of the document and continue discussing this issue, but make it easier for others to also make pull requests if they want.

michalmuskala (approved):
Sounds good to me 👍

PR #128, opened 2020-09-03
erlfmt erlfmt and keep it that way

### LINKED ISSUE #12: erlfmt is not fully erlfmted
How about we format the erlfmt code using erlfmt in the github action and then call `git diff --exit-code .`, which will result in exit code 1, if the pull request is not formatted.

This way we don't regress and keep erlfmt codebase formatted itself.

awalterschulze:
Fixes #12 

Before formatting I only ran this task and got the following output
https://github.com/WhatsApp/erlfmt/runs/1066854528?check_suite_focus=true

Screen Shot 2020-09-03 at 2 37 05 PM

After formatting, checks pass

I had to include removing the generated parser, as this is not committed into the repo and not formatted on generation, so we probably should not check it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,4 +41,4 @@ jobs:
     - name: Release
       run: rebar3 as release escriptize
     - name: Check foramtting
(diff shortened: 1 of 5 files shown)


michalmuskala (approved):
🎉

PR #133, opened 2020-09-08
rather use format_string for checking

awalterschulze:
Changed --check to rather use format_string, this way stdin is also supported and API is simpler.

## Test Plan

```sh
$ rebar3 as release escriptize

$ _build/release/bin/erlfmt --check "{src,include,test}/*.{hrl,erl}" "rebar.config"
Checking formatting...
[warn] src/erlfmt_cli.erl
[warn] src/erlfmt_scan.erl
[warn] test/erlfmt_SUITE.erl
[warn] Code style issues found in the above file(s). Forgot to run erlfmt?

$ _build/release/bin/erlfmt -w "{src,include,test}/*.{hrl,erl}" "rebar.config"
$ git diff
...
modified:   src/erlfmt_cli.erl
modified:   src/erlfmt_scan.erl
modified:   test/erlfmt_SUITE.erl
...

$ _build/release/bin/erlfmt --check "{src,include,test}/*.{hrl,erl}" "rebar.config"
Checking formatting...
All matched files use erlfmt code style!
```

### DIFF
--- a/src/erlfmt_cli.erl
+++ b/src/erlfmt_cli.erl
@@ -31,7 +31,7 @@ opts() ->
         {write, $w, "write", undefined, "modify formatted files in place"},
         {out, $o, "out", binary, "output directory"},
         {verbose, undefined, "verbose", undefined, "include debug output"},
-        {check, undefined, "check", undefined,
+        {check, $c, "check", undefined,
             "Check if your files are formatted."
             "Get exit code 1, if some files are not formatted."
             "--write is not supported."},
--- a/src/erlfmt_cli.erl
+++ b/src/erlfmt_cli.erl
@@ -17,6 +17,7 @@
 
 -record(config, {
     verbose = false :: boolean(),
+    check = false :: boolean(),
     width = undefined :: undefined | pos_integer(),
     pragma = ignore :: erlfmt:pragma(),
     out = standard_out :: erlfmt:out()
@@ -30,10 +31,10 @@ opts() ->
         {write, $w, "write", undefined, "modify formatted files in place"},
         {out, $o, "out", binary, "output directory"},
         {verbose, undefined, "verbose", undefined, "include debug output"},
-        {check, $c, "check", undefined,
+        {check, undefined, "check", undefined,
             "Check if your files are formatted."
             "Get exit code 1, if some files are not formatted."
-            "stdin is not supported and --write is not supported."},
+            "--write is not supported."},
         {print_width, undefined, "print-width", integer,
             "The line length that formatter would wrap on"},
         {require_pragma, undefined, "require-pragma", undefined,
@@ -60,13 +61,22 @@ do(Opts, Name) ->
 do_unprotected(Opts, Name) ->
     case parse_opts(Opts, Name, [], #config{}) of
         {format, Files, Config} ->
-            case Config#config.out of
-                check -> io:format(standard_error, "Checking formatting...~n", []);
+            case Config#config.check of
+                true -> io:format(standard_error, "Checking formatting...~n", []);
                 _ -> ok
             end,
             case parallel(fun(File) -> format_file(File, Config) end, Files) of
                 ok ->
-                    ok;
+                    case Config#config.check of
+                        true ->
+                            io:format(
+                                standard_error,
+                                "All matched files use erlfmt code style!~n",
+                                []
+                            );
+                        _ ->
+                            ok
+                    end;
                 warn ->
                     io:format(
                         standard_error,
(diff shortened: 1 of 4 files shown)


src/erlfmt_cli.erl
```diff
@@ -83,18 +93,48 @@ do_unprotected(Opts, Name) ->
             erlang:halt(2)
     end.
 
+-dialyzer({no_improper_lists, [read_stdin/1]}).
+
+read_stdin(Data) ->
+    case io:get_chars(standard_io, "", 4096) of
+        MoreData when is_binary(MoreData) -> read_stdin([Data | MoreData]);
+        eof -> {ok, Data};
+        {error, Reason} -> throw({error, Reason})
```
michalmuskala, line 102: Do we handle throws nicely in this context? We handle them in `erlfmt`, but I'm not sure here

src/erlfmt_cli.erl
```diff
@@ -83,18 +93,48 @@ do_unprotected(Opts, Name) ->
...
+                Original = unicode:characters_to_list(OriginalBin),
+                case erlfmt:format_string(Original, Options) of
+                    {ok, Formatted, FormatWarnings} ->
+                        case Formatted of
+                            Original -> {ok, FormatWarnings};
+                            _ -> {warn, FormatWarnings}
+                        end;
+                    Other ->
+                        Other
+                end;
```
michalmuskala, line 129: Let's extract this to a function

src/erlfmt_cli.erl
```diff
@@ -139,10 +179,8 @@ parse_opts([insert_pragma | Rest], Name, Files, Config) ->
     parse_opts(Rest, Name, Files, Config#config{pragma = insert});
 parse_opts([{files, NewFiles} | Rest], Name, Files0, Config) ->
     parse_opts(Rest, Name, expand_files(NewFiles, Files0), Config);
-parse_opts([], _Name, [stdin], #config{out = Out}) when Out =/= standard_out ->
-    {error, "stdin mode can't be combined with out or write options"};
-parse_opts([], _Name, [stdin], #config{out = check}) ->
-    {error, "stdin mode can't be combined check mode"};
+parse_opts([], _Name, [stdin], #config{out = Out}) when Out =/= standard_out, Out =/= check ->
```
michalmuskala, line 182: `out` can't be `check` now, right?

src/erlfmt_scan.erl
```diff
@@ -136,6 +136,8 @@ continue(Scan, Inner0, Loc0, Buffer0) ->
 read_rest(#state{inner = undefined}) ->
     %% reached EOF, no further nodes
     {ok, ""};
+read_rest(#state{scan = _Scan, inner = eof, loc = _Loc, buffer = Buffer}) ->
```
michalmuskala, line 139: When does this happen?

src/erlfmt_cli.erl
michalmuskala: Also, does this compose well with `require-pragma`?

src/erlfmt_cli.erl
```diff
@@ -30,10 +31,10 @@ opts() ->
         {write, $w, "write", undefined, "modify formatted files in place"},
         {out, $o, "out", binary, "output directory"},
         {verbose, undefined, "verbose", undefined, "include debug output"},
-        {check, $c, "check", undefined,
+        {check, undefined, "check", undefined,
```
michalmuskala, line 34: Why this change?

src/erlfmt_cli.erl
awalterschulze: do calls and catch throws from do_unprotected which calls format_file, so I think we are fine.
write_formatted also contained throws which we didn't try to catch here.

src/erlfmt_cli.erl
awalterschulze: Great idea, I extracted it.
I also added a test for require-pragma

src/erlfmt_scan.erl
awalterschulze: This was for no_pragma.erl, which has only one line.
We read a single form to check for pragma, so when we call read_rest, the only line has already been read.

src/erlfmt_cli.erl
awalterschulze: I don't know, it just seems more symmetric with --verbose

src/erlfmt_cli.erl
awalterschulze: Good catch. Fixed :)

awalterschulze:
This newest commit changes the API and factors out the writing of the file to erlfmt_cli.
Now check also doesn't need to read the whole string, if it is just going to skip the file.
The format_string and format_file API functions are now even more symmetric.

src/erlfmt_cli.erl
michalmuskala: `--verbose` doesn't use `-v` becuase it's already used for `--version`. We could use `-V` for alias to `--verbose`

awalterschulze: In general, I think it is okay to leave out the short hands for everything that isn't `-h`.
But I am okay with add the `-c` back, but `-V` seems weird.

awalterschulze: Added back -c

src/erlfmt_cli.erl
```diff
@@ -84,38 +94,97 @@ do_unprotected(Opts, Name) ->
...
+        _ ->
+            ok
+    end,
+    case Result of
+        {ok, FormattedText, Warnings} ->
             [print_error_info(Warning) || Warning <- Warnings],
-            ok;
-        {check_failed, _, _, Warnings} ->
+            case Check of
+                true -> ok;
```
michalmuskala, line 120: Do we actually perform the check now? I can't see when it is done

michalmuskala (commented):
I like this approach much more, but I think the `erlfmt_cli` code has a lot of spaghetti conditionals now and I find it rather hard to reason through. Especially the `check` flag is checked again and again. Perhaps there's some better way to do this?

src/erlfmt_cli.erl
```diff
@@ -84,38 +94,97 @@ do_unprotected(Opts, Name) ->
...
             error
     end.
 
+write_formatted(_FileName, Formatted, standard_out) ->
+    io:put_chars(Formatted);
+write_formatted(FileName, Formatted, Out) ->
+    OutFileName = out_file(FileName, Out),
+    case filelib:ensure_dir(OutFileName) of
+        ok -> ok;
+        {error, Reason1} -> throw({error, {OutFileName, 0, file, Reason1}})
```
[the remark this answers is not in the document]
awalterschulze: I changed it to print and not throw

src/erlfmt_cli.erl
awalterschulze: We performed check on stdin and not on files.
I fixed this now
And reran the test plan properly this time.
Good catch, thank you

awalterschulze:
> I like this approach much more, great idea! I think the `erlfmt_cli` code, though, has a lot of spaghetti conditionals now and I find it rather hard to reason through. Especially the `check` flag is checked again and again. Perhaps there's some better way to do this?

I now made `check` part of `out()` again and so removed one level of nesting.

michalmuskala (approved):

PR #140, opened 2020-09-13
Relax map syntax

### LINKED ISSUE #137: Error on macro definition
I know, the code looks super weird but for some reason I ended up with the code and `erlfmt` complains about it :(
Here's a code
```erl
-module(foo).

-export([foo/0]).

-define(FOO,
    foo => bar
).

foo() ->
    #{
        ?FOO
    }.
```

and error message from `rebar3 fmt`

```
src/foo.erl:5:9: syntax error before: '=>'
src/foo.erl:11:5: syntax error before: '}'
```

BTW thanks for sane code formatter for Erlang, it was a major missing productivity tool.

michalmuskala:
This changes the `=>` and `:=` operators from special to maps
to generic operators that can form valid expresions anywhere,
and maps to `#{ any_expression... }` instead of only accepting
the map operators.

This allows easily supporting map expressions behind macros.

Closes #137

### DIFF
--- a/src/erlfmt_parse.yrl
+++ b/src/erlfmt_parse.yrl
@@ -28,7 +28,7 @@ list_comprehension lc_expr lc_exprs
 binary_comprehension
 tuple
 record_expr record_name record_field_name record_tuple record_field record_fields
-map_expr map_tuple map_field map_fields map_key
+map_expr map_tuple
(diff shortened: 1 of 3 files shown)


awalterschulze (approved):
Impressive fix :)

PR #145, opened 2020-09-16
add Makefile and use it in github actions

awalterschulze:


ilya-klyuchnikov:
One concern/note/remark - make is not a first-class citizen on windows. - If you can keep the things as cross-platform as possible - it would be perfect.

awalterschulze:
Yes this is just for our personal development.
We are concerned about making this easier for others, but will be tough for us, since we can call erlfmt on ourselves without copying and pasting that line.

Makefile
```diff
@@ -0,0 +1,35 @@
...
+.PHONY: clean
+clean:
+	rm -rf _build
+	rebar3 clean
+
+.PHONY: fmt
+fmt:
+	make clean
+	make release
+	_build/release/bin/erlfmt -w "{src,include,test}/*.{hrl,erl,app.src}" "rebar.config"
```
michalmuskala, line 22: Any reason for not using target dependencies?
```suggestion
fmt: clean release
	_build/release/bin/erlfmt -w "{src,include,test}/*.{hrl,erl,app.src}" "rebar.config"
```

I'm also not sure we do need to rebuild everything every time we compile. Any particular reason for this?

michalmuskala:
@ilya-klyuchnikov I see this as a short-term solution to simplify maintenance. Long-term we need a way to provide config for standalone escript as well, not only for the rebar integration. Additionally running `--check` should automatically pick up files from that config. This should remove the need for any wrappers.

Makefile
awalterschulze: Paranoia and Parity with CI
But I have changed it, so that only the `check` command does `clean` now.

Generally I prefer calling make over target dependencies, for readability, but I have now added target dependencies to conform :)

awalterschulze:
> @ilya-klyuchnikov I see this as a short-term solution to simplify maintenance. Long-term we need a way to provide config for standalone escript as well, not only for the rebar integration. Additionally running `--check` should automatically pick up files from that config. This should remove the need for any wrappers.

Maybe it is time to create this as an issue, so we don't forget?

michalmuskala (approved):

PR #184, opened 2020-09-23
fix issue with comments before a dot

### LINKED ISSUE #135: Comments: Incorrect formatting of type spec
The type spec

```erlang
-type t() ::
    a
    % comment
    .
```

gets formatted as

```erlang
-type t() ::
    a
    % comment.
```

The case gets detected and outputs the following:

```
test.erl:1:1: formatter result not equivalent. Please report this bug.

{attribute,#{end_location => {4,6},location => {1,1}},
           {atom,#{end_location => {1,6},location => {1,2},text => "type"},
                 type},
           [{op,#{end_location => {3,14},
                  inner_end_location => {2,6},
                  location => {1,7},
                  post_comments =>
                      [{comment,#{end_location => {3,14},location => {3,5}},
                                ["% comment"]}]},
                '::',
                {call,#{end_location => {1,10},location => {1,7}},
                      {atom,#{end_location => {1,8},
                              location => {1,7},
                              text => "t"},
                            t},
                      []},
                {atom,#{end_location => {2,6},location => {2,5},text => "a"},
                      a}}]}

{raw_string,#{end_location => {1,14},location => {1,1}},
            "-type t() ::\n    a\n% comment."}
```

awalterschulze:
Fixes #135

### DIFF
--- a/test/erlfmt_SUITE_data/simple_comments.erl
+++ b/test/erlfmt_SUITE_data/simple_comments.erl
@@ -14,9 +14,8 @@
 -define(STR(X),
     %% comment 1
     ??X
-)
-%% comment 2
-.
+    %% comment 2
+).
 
 -define(FUN(M, F), M:F).
 
--- a/test/erlfmt_SUITE_data/simple_comments.erl
+++ b/test/erlfmt_SUITE_data/simple_comments.erl
@@ -14,8 +14,9 @@
 -define(STR(X),
     %% comment 1
     ??X
-    %% comment 2
-).
+)
+%% comment 2
+.
 
 -define(FUN(M, F), M:F).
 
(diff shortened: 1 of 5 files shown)


test/erlfmt_SUITE_data/simple_comments.erl
```diff
@@ -14,8 +14,9 @@
 -define(STR(X),
     %% comment 1
     ??X
-    %% comment 2
-).
+)
+%% comment 2
+.
```
michalmuskala, line 19: Ah... seeing this now I'm less sure this is the correct way to fix. I assume we'll do this in more cases where there's a "trailing" comment at the end of an attribute, but inside parenthesis. 

I'm not sure we can get away without having some separate space to store the range `.` is covering as opposed to the thing inside parentheses.

awalterschulze: Good point.  Thanks for helping to fix it :)

michalmuskala (approved):

PR #185, opened 2020-09-23
Update instructions for integration with rebar and CI

mheiber:
With the old recommendation of rebar config, it was difficult (impossible?) to run `rebar3 fmt --check` in CI and have it check formatting.

This PR changes the recommended setup so that the command users must run in order to format things is `rebar3 fmt --write`, not just `rebar3 fmt`. Then they can do `rebar3 fmt --check` in CI.

I probably have not thought through all the implications or updated all the places that need to be updated, so please consider this a "discussion" PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,6 @@ can be configured with defaults in your `rebar.config`, for example:
 
 ```erlang formatted rebarconfig2
 {erlfmt, [
-    write,
     {files, "{src,include,test}/*.{hrl,erl}"}
 ]}.
 ```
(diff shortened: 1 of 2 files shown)


michalmuskala (approved):
That makes sense to me, at least until we figure out some better interaction between CLI arguments and config in `rebar.config`

awalterschulze (approved):
Yeah looks good to me.
Thank you for also updating the readme :)

michalmuskala:
Thank you! :heart:

PR #187, opened 2020-09-24
formatting decision document: commas in lists

awalterschulze:


elbrujohalcon:
Since codebases are generally either all-comma-first or all-comma-last, by checking 2 of them where comma-last is used you were certainly bound to get the results you've got…
What if, at least for completeness sake you check other codebases, like the repos in https://github.com/klarna and the repos in https://github.com/inaka.
Also, for this exercise or future exercises, another *huge* codebase that you can use for analysis is Kazoo's https://github.com/2600hz (Being myself one of the maintainers of Elvis, I can tell you they were the only devs ever that complained about its speed… their amount of erlang modules is staggering).

awalterschulze:
Good idea, I have add some more data, but it seems to be tough to find large enough and popular enough code bases that counter comma last style.

elbrujohalcon:
Yeah, for Inaka… you have to count on _all of the repos_ (as opposed to just one of them) if you want more data.
Anyway, another big pile of code that's is very popular (according to the number of job offers I get to _fix/install/configure_ it 🤦 ) is [ejabberd](https://github.com/processone/ejabberd) and it's nicer sibling [MongooseIM](https://github.com/esl/MongooseIM).

awalterschulze:
Thank you I added those data sets too.
If you give me a bash script I can run to clone a but of Inaka repos, I can add that too.

elbrujohalcon:
Would a repo containing all of them as dependencies be good enough?
You can run `rebar3 as test compile` on it and then analyze everything under `_build`.
If that's so… I believe this one is the one with the largest tree: https://github.com/inaka/lsl

awalterschulze:
I tried that, but I get 
```
===> /Users/walterschulze/code/src/github.com/awalterschulze/lsl/c_src/bcrypt_port.c:20:10: fatal error: 'erl_interface.h' file not found
#include <erl_interface.h>
         ^~~~~~~~~~~~~~~~~
1 error generated.
```

I am also worried about pulling in more dependencies than just inaka.
It would be nice if this was something more reproducible, if that makes sense.

elbrujohalcon:
This should do it, then.
[clone-all.zip](https://github.com/WhatsApp/erlfmt/files/5282669/clone-all.zip)

awalterschulze:
Thank you I have add inaka to the document.

doc/FormattingDecisionCommas.md
```diff
@@ -0,0 +1,185 @@
...
+
+[MongooseIM](https://github.com/esl/MongooseIM):
+  - Commas as a Suffix: 852
+  - Commas as a Prefix: 1
+
+[ejabberd](https://github.com/processone/ejabberd):
+  - Commas as a Suffix: 173
+  - Commas as a Prefix: 0
+
+[inaka repos](#inaka)
```
elbrujohalcon, line 123:
```suggestion
[inaka repos](#inaka):
```

elbrujohalcon:
Now… if you want to also analyze a _whole new style_ of coding, you should check the works of @robertcarbone (like [wxi](https://github.com/RobertCarbone/wxi)), which include some jewels like [this](https://github.com/RobertCarbone/wxi/blob/f96ba31c9759e4a872284ae780ce8588727e495c/examples/wxicalc.erl#L162-L167)

```erlang
readdisp(S = #calcst{}) ->
    {ok, [Disp], _} = io_lib:fread("~f", S#calcst.disp ++ case S#calcst.dot of
               true -> "0";
               false -> ".0"
           end),
    Disp.
```

elbrujohalcon:
[These](https://scriptculture.com/#u107259-4) were the examples I was actually looking for…

```erlang
-export([
                           getId/1,
                     getSkipped/1,
                  getTimestamp/1,
               isCommandEvent/1,
             stopPropagation/1,
          resumePropagation/2,
           shouldPropagate/1,
                     skip/1,
                    skip/2

                             ]).
```

awalterschulze:
> [These](https://scriptculture.com/#u107259-4) were the examples I was actually looking for…
> 
> ```erlang
> -export([
>                            getId/1,
>                      getSkipped/1,
>                   getTimestamp/1,
>                isCommandEvent/1,
>              stopPropagation/1,
>           resumePropagation/2,
>            shouldPropagate/1,
>                      skip/1,
>                     skip/2
> 
>                              ]).
> ```

This is what `%% erlfmt-ignore` was built for https://github.com/WhatsApp/erlfmt/blob/master/doc/FormattingDecisionIgnore.md
Actually it might be a nice example to contribute to that document, if you are interested.

elbrujohalcon:
Well… it's not _my_ code, you should ask @robertcarbone about it, but I guess he'll be honored.

awalterschulze:
I think this pull request is ready to merge now

paulo-ferraz-oliveira:
@awalterschulze: you briefly mentioned `%% erlfmt-ignore`. What is it? How does it work?

awalterschulze:
@paulo-ferraz-oliveira see the [readme](https://github.com/WhatsApp/erlfmt#ignoring-formatting)
It helps you to skip formatting of specific top level expressions.

paulo-ferraz-oliveira:
> I do not know if it will work for all spec use cases, but some should be covered.

I was kidding about the specs. I think that in general the formatter decisions are OK. I'll just have to ponder on some of them some more. In any case, good to know there's an option for when I want to put ASCII art in my code 😄.

> So please be patient.

😄 no hurry here. Curiosity, but not pressure.

awalterschulze:
I know, but sometimes jokes also deserve a serious answer ;)

Case and point:
For ascii art you can still use comments, because we don't touch the spacing inside those.

paulo-ferraz-oliveira:
> I know, but sometimes jokes also deserve a serious answer ;)
> 
> Case and point:
> For ascii art you can still use comments, because we don't touch the spacing inside those.

Ah, no. I'm talking about the juicy stuff, like https://ferd.ca/obfuscation.html.

awalterschulze:
> > I know, but sometimes jokes also deserve a serious answer ;)
> > Case and point:
> > For ascii art you can still use comments, because we don't touch the spacing inside those.
> 
> Ah, no. I'm talking about the juicy stuff, like https://ferd.ca/obfuscation.html.

Ha ha :D
Obfuscated code is not meant to be read, so I suspect it is not important to format ;P
But you could use the `--require-pragma` flag and not add an `%% @format` comment to that file.

michalmuskala (approved):

PR #188, opened 2020-09-24
parse attributes with empty parens

### LINKED ISSUE #178: ifdef: Empty line before `-endif`
It seems that before `-endif`, an empty line is being forced - I imagine it's because the previous function "closes".

For consistency (if it's not a bug) shouldn't the same be done for `-ifdef` (et al.) - add an empty line after it?

awalterschulze:
Fixes #178

attributes can have zero parameters, for example:
```erlang
-endif().
```

For more details on how this was discovered, see the issue


paulo-ferraz-oliveira:
😃 I hadn't even noticed those `()`, to be honest. I don't use them, by default, so it must've been a moment of distraction. Nice catch!

awalterschulze:
It took me a while to see it too :)
But either way, it is valid so we should handle it.

Thanks again for reporting it.
That very useful.
Nice catch yourself :D

PR #192, opened 2020-09-25
formatting decision document: number of spaces

awalterschulze:


doc/FormattingDecisionSpaces.md
```diff
@@ -0,0 +1,93 @@
...
+  - 2 spaces: 30
+  - 4 spaces: 13740
+
+[ejabberd](https://github.com/processone/ejabberd):
+  - 2 spaces: 349
+  - 4 spaces: 9010
+
+[Inaka repos](#inaka):
+  - 2 spaces: 2800
+  - 4 spaces: 1349
```
paulo-ferraz-oliveira, line 45: Hm... in `elvis_core` alone I find 4329 matches (`master`).

paulo-ferraz-oliveira: And if I count only 4-space blocks alone, it jumps to 9725 (but this is probably not the same measure).

awalterschulze: Hmmm did you run the same `pcregrep -Mr` regex or did you use another tool?

awalterschulze: Also did you remove your _build folders?

paulo-ferraz-oliveira: I didn't use the `pcregrep`. Let me check that one.

paulo-ferraz-oliveira: Ah, I see, you're counting indentation only "if after `->`". That's fair, I guess.

awalterschulze: Yes I needed something that would not include two x 2 spaces being confused with 4 spaces.  So there are more scenarios for that, but I thought `->` is the easiest and has least chance for confusion.

doc/FormattingDecisionSpaces.md
```diff
@@ -0,0 +1,93 @@
...
+$ pcregrep -Mr --include=".*\.erl" --include=".*\.hrl" "\->\n\s\s\s\s\w" . | grep "\->" | wc -l
+# count 2 spaces
+$ pcregrep -Mr --include=".*\.erl" --include=".*\.hrl" "\->\n\s\s\w" . | grep "\->" | wc -l
+```
+
+## Results
+
+OTP:
+  - 2 spaces: 16366
+  - 4 spaces: 113481
```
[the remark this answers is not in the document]
awalterschulze: Very interesting.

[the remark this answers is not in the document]
elbrujohalcon: Only [very recently](https://github.com/erlang/otp/commit/5b6351fdd10c7f97d414f2e988d4a6f06d7de598) they changed that in the pretty-printer, actually.

But there are still [traces of the behavior](https://github.com/erlang/otp/blob/8ef1bc22b940c085be26024af090f1eeec71eaed/lib/syntax_tools/src/prettypr.erl#L1288-L1303).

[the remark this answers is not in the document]
awalterschulze: Nice find.
I guess we should also consider 8 spaces in the data analysis then.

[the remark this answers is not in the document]
elbrujohalcon: You won't find it, tho… Their algorithm was "indent some stuff with 2 spaces and some stuff with 4 spaces, but if the indentation covers more than 8 spaces, then replace those 8 spaces with a tab".
That means that if you only check indentation _after a `->`_, you'll only find 4-space indentation. To find lines indented with 2 spaces you need to find `case` statements whose expression spans multiple lines (among other similar scenarios). And to find 8-space indentation, you need 2 4-space indentations (or one 4-space and 2 2-space indentations) combined.

michalmuskala (approved):

doc/FormattingDecisionSpaces.md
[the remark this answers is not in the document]
KennethL: OTP use 4 spaces for indentation (there are some cases where half indent is used (2 spaces). It is the emacs mode that is considered the norm. With these settings:

    (setq-default indent-tabs-mode nil)
    (setq c-basic-offset 4)

Previously if there where 8 spaces or more in a row , 8 spaces was converted to a tab, but not any more.
Old code has not been reformatted so it may contain tabs.

The half indent is used for example before `when` like this:

    foo(A) 
      when is_integer(A) ->
        A+1.

But of course you should not write like that if everything including `->` fit in the same line.

So from OTP point I would say 4 spaces is the norm for indentation (absolutely not 8 spaces).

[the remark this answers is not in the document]
awalterschulze: Phew, Thank you that saves me the work of extra analysis :)

PR #197, opened 2020-09-30
Recommend to use project_plugins instead of plugins in readme.

michalmuskala:
plugins are also downloaded if the project is a transitive dependency,
project_plugins are downloaded only when working directly on the codebase.
For the tasks erlfmt is used for normally, project_plugins seems more
appropriate

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ The easiest way to use erlfmt is as a rebar plugin, by adding to your
 `rebar.config`:
 
 ```erlang formatted rebarconfig1
-{plugins, [erlfmt]}.
+{project_plugins, [erlfmt]}.
 ```
 
 This will provide a new `rebar3 fmt` task. All erlfmt command-line options


awalterschulze:
Should we change this in our rebar.config file too?

paulo-ferraz-oliveira:
@awalterschulze: it's not a big deal unless people import `erlfmt` as a dependency, which I guess is not the usual use case.

awalterschulze (approved):

michalmuskala:
We don't have this in our `rebar.config` since rebar doesn't really support using the project on itself as a plugin

paulo-ferraz-oliveira:
> We don't have this in our `rebar.config` since rebar doesn't really support using the project on itself as a plugin

That would be nice though, wouldn't it? I tend to do it like so:
1. `rebar3 escriptize`,
2. `_build/.../erlfmt` + options.

That way you run the plugin on your own code.

awalterschulze:
It would be really nice, but locally too, not just on travis/github actions.
Although I am happy to be considered an edge case, given this is a language tool.
Nothing a Makefile on top cannot solve :)

paulo-ferraz-oliveira:
For `erlfmt` it wouldn't make (a lot of) sense to have this in `.travis.yml`, unless you had a return code not-0 every time you actually changed a file; and the build would fail in that case 😄.

In `elvis` case it does make sense, though, since it's an analysis tool.

If you don't want to go with a Makefile, you can take a look at [rebar_cmd](https://github.com/gootik/rebar_cmd), for example.

paulo-ferraz-oliveira:
I didn't read the "just" word. I read "It would be really nice, but locally too, not ~just~ on travis/github actions." 😄

Issue #149, opened 2020-09-17
List comprehensions: Inconsistent indentation in list comprehension


david-kubecka:
```
[
    "\"" ++ binary_to_list(A2) ++ "\" -> \"" ++ binary_to_list(A1) ++ "\" " ++ graphviz_edge_label(L)
    || {_, A1, A2, L} <- Es
]
```
gets formatted into
```
    [
        "\"" ++
            binary_to_list(A2) ++
            "\" -> \"" ++ binary_to_list(A1) ++ "\" " ++ graphviz_edge_label(L)
        || {_, A1, A2, L} <- Es
    ]
```
I would expect something like
```
    [
        "\"" ++
            binary_to_list(A2) ++
            "\" -> \"" ++
            binary_to_list(A1) ++
            "\" " ++ graphviz_edge_label(L)
        || {_, A1, A2, L} <- Es
    ]
```
with similar reasoning as in https://github.com/WhatsApp/erlfmt/issues/136#issuecomment-691654159.

Ideally it should like
```
    [
        "\"" ++
            binary_to_list(A2) ++
            "\" -> \"" ++
            binary_to_list(A1) ++
            "\" " ++
            graphviz_edge_label(L)
        || {_, A1, A2, L} <- Es
    ]
```
Note the final `++` argument is on an extra line.

awalterschulze:
Thank you @david-kubecka for the example.  This is great feedback.
I think this is similar to https://github.com/WhatsApp/erlfmt/issues/136
I will take a look.

awalterschulze:
In the interim you can try to add a newline after the final `++` and check if `erlfmt` preserves it for you.

michalmuskala:
Yes, this is exactly the same issue as #136, but in a different context. We should merge them.

awalterschulze:
Yes I agree, I have copied the example to the duplicate issue https://github.com/WhatsApp/erlfmt/issues/136
Closing this.
I hope that makes sense.

PR #198, opened 2020-10-02
consistent binary operator breaking

### LINKED ISSUE #136: Binary Operators: Last binary operator not formatted consistently
Hi. I tried to run it on my pet-project. Need to get used to a new style now :-) Looks great overall, thanks!
Found some strange example:

```diff
@@ -88,9 +100,16 @@ text_spec2char_ratio(Bin) ->
     NSpec = s2c(Bin, 0),
     NSpec / Size * 100.
 
-s2c(<<C, Rest/binary>>, Acc) when C == $\s orelse C == $\t orelse C == $\r orelse C == $\n
-                                  orelse C == $\;  orelse C == $\$ orelse C == $\>
-                                  orelse C == $\< orelse C == $\/ ->
+s2c(<<C, Rest/binary>>, Acc) when
+    C == $\s orelse
+        C == $\t orelse
+        C == $\r orelse
+        C == $\n orelse
+        C == $\; orelse
+        C == $\$ orelse
+        C == $\> orelse
+        C == $\< orelse C == $\/
+->
     s2c(Rest, Acc + 1);
 s2c(<<_C/utf8, Rest/binary>>, Acc) ->
     s2c(Rest, Acc);
```
* Why all the `orelse`s are on individual lines, but last one has 2 `C == $\< orelse C == $\/`. It looks inconsistent.
* Why single `->` on a newline?
* Minor: 1st `orelse` clause: `C == $\s orelse` has different indentation than others (guess most of other editors do the same, but I never understood why)

awalterschulze:
Fixes #136 and #149

```erlang
Foo orelse Bar orelse Baz
```

will now break to

```erlang
Foo orelse
    Bar orelse
    Baz
```

Instead of

```erlang
Foo orelse
    Bar orelse Baz
```

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -307,13 +307,15 @@ format_nodes_loop(
 ) when Name =:= 'if'; Name =:= 'ifdef'; Name =:= 'ifndef'; Name =:= 'else' ->
     % preserve empty line after
     [$\n, format_node(Attr, PageWidth)] ++
-        maybe_empty_line(Attr, NextNode) ++ format_nodes_loop(Rest, PageWidth);
+        maybe_empty_line(Attr, NextNode) ++
+        format_nodes_loop(Rest, PageWidth);
 format_nodes_loop([Node | [{attribute, _, {atom, _, Name}, _} = Attr | _] = Rest], PageWidth) when
     Name =:= 'else'; Name =:= 'endif'
 ->
     % preserve empty line before
     [$\n, format_node(Node, PageWidth)] ++
-        maybe_empty_line(Node, Attr) ++ format_nodes_loop(Rest, PageWidth);
+        maybe_empty_line(Node, Attr) ++
+        format_nodes_loop(Rest, PageWidth);
 format_nodes_loop([{attribute, _, {atom, _, RepeatedName}, _} | _] = Nodes, PageWidth) ->
     {Attrs, Rest} = split_attrs(RepeatedName, Nodes),
     MaybeEmptyLine =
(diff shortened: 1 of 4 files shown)


src/erlfmt.erl
```diff
@@ -409,7 +411,9 @@ equivalent({Type, _, L1, L2, L3}, {Type, _, R1, R2, R3}) ->
     equivalent(L1, R1) andalso equivalent(L2, R2) andalso equivalent(L3, R3);
 equivalent({Type, _, L1, L2, L3, L4}, {Type, _, R1, R2, R3, R4}) ->
     equivalent(L1, R1) andalso
-        equivalent(L2, R2) andalso equivalent(L3, R3) andalso equivalent(L4, R4);
+        equivalent(L2, R2) andalso
+        equivalent(L3, R3) andalso
+        equivalent(L4, R4);
```
michalmuskala, line 416: Looking at this change specifically, I'm actually not 100% convinced it's better now compared to what it was. I think we should reconsider how we handle things like this.

awalterschulze: I had a similar initial reaction, but I have being growing more and more fond of this solution after looking more and more at it and especially when considering our current inconsistent alternative.

Also one way to avoid this is for the user to group parts of the expression in parens to keep it on one line

awalterschulze: @michalmuskala proposed trying to rotate the associativity, so that this would break as:
```
equivalent(L1, R1) andalso equivalent(L2, R2) andalso equivalent(L3, R3) andalso
         equivalent(L4, R4)
```
which we hope is what most people would expect

awalterschulze: And I think this is worth trying

elbrujohalcon: Don't forget to test it with stuff like…

```erlang
AVariable andalso AnotherVariable andalso AnotherVariableMore …
```

```erlang
a_function:with_a(<<"very">>, Long, list, "of", <<"attributes">>) andalso a:short(one) andalso another:very_very_long_one() andalso …
```

```erlang
the:ever(funny) andalso begin blocks:of(code), that:appear(between), your:boolean(expressions) end andalso are:terrible(things, to, debug)
```

…and of course… [else-less ifs](https://medium.com/erlang-battleground/if-i-fell-in-love-with-you-f35cf6b80d15)… 🤦‍♂️ 
```erlang
is_registered(User)
  andalso password_matches(User, Password)
  andalso (has_active_session(User) orelse is_admin(User))
  andalso
    begin
      register_login(User, dates:now()),
      [ open_admin_session(User) || is_admin(User) ]
    end
```

awalterschulze: cool thank you for the test cases :)

awalterschulze:
Superceded by https://github.com/WhatsApp/erlfmt/pull/262

PR #201, opened 2020-10-02
fix record field type end location

### LINKED ISSUE #142: Records: Somewhat strange format if default in record is multiline
This seems like an edge-case where the default is multiline, that I don't think we carefully considered how it should look like. 

For example, it formats like this today:
```erlang
-record(foo, {
    field = #{
        key1 => [],
        key2 => []
    } ::
        type()
}).
```

Perhaps something like the following would be better:
```erlang
-record(foo, {
    field = #{
        key1 => [],
        key2 => []
    } :: type()
}).
```

awalterschulze:
Fixes #142


michalmuskala (approved):
Great catch and a very subtle bug

PR #204, opened 2020-10-05
provide default value for --files when none is provided

### LINKED ISSUE #169: rebar3: Crash upon `rebar3 fmt **`
I'm sure this syntax is wrong, but it was the first thing I tried when I ran `rebar3 fmt` and was told about `no files provided to format`.

**What was I expecting?**

As per [filelib:wildcard/1](https://erlang.org/doc/man/filelib.html#wildcard-1), to "match all files and zero or more directories and subdirectories."

Using `master`.

By the way:

* `/**`
* `src/**`

also crash.

**What was I expecting?**

That the formatter would execute, and if files had to be skipped, they would be - with a warning, stating that it can't format so and so, because "unknown file type" -, instead of crashing 😄 

Oh, I now see that it's not expected to crash, since it's in the `README.md`. Anyway, here's my report...

awalterschulze:
Fixes #169

### DIFF
--- a/src/erlfmt_cli.erl
+++ b/src/erlfmt_cli.erl
@@ -35,8 +35,8 @@ opts() ->
         {out, $o, "out", binary, "output directory"},
         {verbose, undefined, "verbose", undefined, "include debug output"},
         {check, $c, "check", undefined,
-            "Check if your files are formatted."
-            "Get exit code 1, if some files are not formatted."
+            "Check if your files are formatted. "
+            "Get exit code 1, if some files are not formatted. "
             "--write is not supported."},
         {print_width, undefined, "print-width", integer,
             "The line length that formatter would wrap on"},
@@ -44,12 +44,13 @@ opts() ->
             "Require a special comment @format, called a pragma, "
             "to be present in the file's first docblock comment in order for prettier to format it."},
         {insert_pragma, undefined, "insert-pragma", undefined,
-            "Insert a @format pragma to the top of formatted files when pragma is absent."
-            "Works well when used in tandem with --require-pragma, but"
-            "it is not allowed to use require-pragma and insert-pragma at the same time."},
+            "Insert a @format pragma to the top of formatted files when pragma is absent. "
+            "Works well when used in tandem with --require-pragma, "
+            "but it is not allowed to use require-pragma and insert-pragma at the same time."},
         {files, undefined, undefined, string,
-            "files to format, - for stdin."
-            "If no files are provided then {src,include,test}/*.{hrl,erl,app.src} and rebar.config"
+            "files to format, - for stdin. "
+            "If no files are provided and one option of [-w, -o, -c] are provided then "
+            "{src,include,test}/*.{hrl,erl,app.src} and rebar.config "
             "are specified as the default."}
     ].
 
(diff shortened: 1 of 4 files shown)


src/erlfmt_cli.erl
```diff
@@ -58,7 +61,23 @@ do(Name, Opts) ->
 do(Name, PreferOpts, DefaultOpts) ->
     PreferParsed = parse_opts(PreferOpts),
     DefaultParsed = parse_opts(DefaultOpts),
-    Parsed = resolve_parsed(PreferParsed, DefaultParsed),
+    case {specified_files(PreferOpts), PreferParsed} of
+        {SpecifiedFiles, {format, [], _}} when SpecifiedFiles =/= [] ->
```
michalmuskala, line 65: If `PreferParsed` returns an empty list of files to format, do we need to run the `specified_files` check? What's the case when they would return different values?

michalmuskala:
What's the behaviour when you just run `erlfmt` without any options? Today it prints the help, will it now format things to stdout? I think it would be worth it to preserve the current behaviour. We could provide the default files when one of the "action" flags is passed to erlfmt, like `-w`, `-c`, or `-o`

src/erlfmt_cli.erl
awalterschulze: If you do not specify files, we are going to add the default files, but we do not want to add these default files in cases where you specified files that just didn't match any existing files.

awalterschulze:
Yes it prints to stdout.  I think writing files to stdout is harmless and we do provide classic `-h` and `--help` flags for help.
We could enable this only for those specific flags, but then there is no way to invoke the stdout behaviour for the default files.

michalmuskala:
I would argue stdout with multiple files is quite useless in the first place - you can't tell where one file ends, the other starts, similar to `cat`ing multiple files together.

awalterschulze:
Cool sounds good, I will update it to only add default files, given at least one other parameter.

paulo-ferraz-oliveira:
Would PR title "provide default value for --files when none is provided" be clearer? (by reading it, I'm under the impression actual files are going to be added) 😄

paulo-ferraz-oliveira:
The README still says `rebar3 fmt` will "format all the files in your project" (it doesn't), but my main issue is that `rebar3 fmt src/*.erl` doesn't work. Isn't this syntax OK? (I'm imagining you use `wildcard` behind the scenes)

michalmuskala (approved):

awalterschulze:
I have now also tested `rebar3 fmt src/*.erl` and it formats my .erl files in my local example project.

awalterschulze:
I have also checked the Readme and it says "**Now** you can format all the files in your project by running:" which implies that you have setup the rebar config.

awalterschulze:
@ilya-klyuchnikov I think you also wanted this feature.

michalmuskala:
There are two layers of file expansion.

When you call `rebar3 fmt src/*.erl` it's the shell that expands the files. When you call `rebar3 fmt 'src/*.erl'` it's erlfmt using `filelib:wildcard/1` - this is generally needed since windows doesn't do wildcard expansion in the shell and we also want to support some convenient CLI for users there. `filelib:wildcard/1` is also used for config from `rebar.config`.

paulo-ferraz-oliveira:
Thanks for the explanation @michalmuskala. I understand the reasoning now, but having to use `'` is still surprising 😄 (it worked though).

awalterschulze:
@michalmuskala wow I never knew that.  Quite embarrassing. 
I had to check and found that the behaviour is the same in Go.  I think I avoided asking this question, because I usually used Go's special `...` syntax, which doesn't get expanded by unix and then we could expand it with a function call.
I always thought the `...` was weirdly inconsistent, but now I get that, that was on purpose to avoid the unix expansion.

PR #206, opened 2020-10-05
bind pipe tighter

### LINKED ISSUE #181: Maps: Curious formatting of alternatives in type spec for map field
(erlfmt 0.7)

The following type spec for a map argument has a `name` field with three alternative types, but erlfmt only indents the first alternative:

```erlang
-spec child_spec(#{
    name =>
        {local, Name :: atom()} |
    {global, GlobalName :: any()} |
    {via, Module :: atom(), ViaName :: any()},
    another_field => atom()
}) -> supervisor:child_spec().
```

If I join all the four lines to make it `name => {local, Name :: atom()} | {global, GlobalName :: any()} | {via, Module :: atom(), ViaName :: any()}` and reformat, it splits that line, puts the first alternative on the first line and the rest on new lines – but without indentation:

```erlang
-spec child_spec(#{
    name => {local, Name :: atom()} |
    {global, GlobalName :: any()} |
    {via, Module :: atom(), ViaName :: any()},
    another_field => atom()
}) -> supervisor:child_spec().
```

I'm not sure what the correct behaviour here would be (but the first case above seems wrong).

Maybe something like this makes sense when the cases need to be split over multiple lines?

```erlang
-spec child_spec(#{
    name =>
        {local, Name :: atom()} |
        {global, GlobalName :: any()} |
        {via, Module :: atom(), ViaName :: any()},
    another_field => atom()
}) -> supervisor:child_spec().
```

awalterschulze:
Fixes #181

When debugging 181 I found that 
```erlang
name =>
        {local, Name :: atom()} |
    {global, GlobalName :: any()} |
    {via, Module :: atom(), ViaName :: any()}
```

Was parsed as
```
name =>
        {local, Name :: atom()}
```
and then the rest, but the pipe needed to bind tighter than the `=>` operator.

### DIFF
--- a/src/erlfmt_parse.yrl
+++ b/src/erlfmt_parse.yrl
@@ -86,9 +86,9 @@ Nonassoc 800 ':='.
 %% Types
 
 Right 40 '::'.
-Right 170 '|'.
-Nonassoc 200 '..'.
-Nonassoc 200 '*'. % for binary expressions
+Right 900 '|'.
+Nonassoc 1000 '..'.
+Nonassoc 1000 '*'. % for binary expressions
 
 node -> attribute dot : setelement(2, '$1', ?range_anno('$1', '$2')).
 node -> function dot : setelement(2, '$1', ?range_anno('$1', '$2')).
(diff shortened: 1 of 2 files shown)


michalmuskala (approved):
Good catch!

PR #207, opened 2020-10-06
Comments below shebang are pull up towards it

### LINKED ISSUE #205: Formatting an escript may disable emulator directives
Consider the following script, which simply prints the cookie of the node (`my_cookie`)

```erlang
#!/usr/bin/env escript
%% -*- erlang -*-
%%! -smp enable -sname foo -setcookie my_cookie

main(_) ->
    io:format("My cookie: ~s~n", [erlang:get_cookie()]).
```

After being formatted with `erlfmt` (version `0.7.0`), a blank line will be inserted between the first and second lines, and as such, the emulator arguments will be ignored. As such, the script won't start with `smp`, and the cookie will be `nocookie`. 

```erlang
#!/usr/bin/env escript

%% -*- erlang -*-
%%! -smp enable -sname foo -setcookie my_cookie

main(_) ->
    io:format("My cookie: ~s~n", [erlang:get_cookie()]).
```

From the [escript docs](https://erlang.org/doc/man/escript.html): 

> On the third line (or second line depending on the presence of the Emacs directive), arguments can be specified to the emulator

As such, the emulator directive line needs to _always_ be the second or third line of the escript file for it to be effective.

Bear in mind that if we remove the Emacs directive from the original script, the formatted version still works, because the blank line leaves the emulator directive on the 3rd line:

```erlang
#!/usr/bin/env escript

%%! -smp enable -sname foo -setcookie my_cookie

main(_) ->
    io:format("My cookie: ~s~n", [erlang:get_cookie()]).
```

awalterschulze:
Fixes #205


src/erlfmt_scan.erl
```diff
@@ -178,12 +179,12 @@ last_node_string(#state{original = [First | _] = Tokens}) ->
     Location = erl_scan:location(First),
     {String, token_anno([{text, String}, {location, Location}])}.
 
+is_white_space({white_space, _, [$\n | _]}) -> true;
+is_white_space(_) -> false.
```
michalmuskala, line 183: This checks for newline specifically, not just any whitespace - this name is not that accurate

michalmuskala (approved):
I think this is a good solution. When looking at the PR, I had another thought that we could solve it another way:

If we see the shebang, we hard-code collecting comments from the next 2 lines and attach them as post comments to the shebang node, instead of what would happen normally and attaching them as pre-comments to the next node. This seems a bit more accurate semantically to what's happening, though the outcome would be exactly the same.

src/erlfmt_scan.erl
awalterschulze: Yes good catch.  I will change this.

awalterschulze:
Your solution sounds much better, because mine is quite hacky.
Let me see if I can try to implement your idea.

awalterschulze:
I had to export comments_with_dot, because we forgot about the case where we could have

```
%% @format

-module(mymodule)
%% some comment
.
```

awalterschulze:
Rereviewed together :)

PR #209, opened 2020-10-14
Warn on long lines

### LINKED ISSUE #120: Warnings: Warn about overlong lines?
In some cases, erlfmt produces lines longer than the print-width (this could be inevitable, for example with very long identifiers).
Would it be desirable to issue a warning about such lines?

ampleyfly:
Closes #120

Use a map in the return values of the format functions, so that more return data can be added.
Add "verbose warnings", which are only printed if the verbose option is given.

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -100,7 +100,12 @@ format_string(String, Options) ->
                     end,
                 Formatted = format_nodes(NodesWithPragma, PrintWidth),
                 verify_nodes("nofile", NodesWithPragma, Formatted),
-                {ok, unicode:characters_to_list(Formatted), Warnings};
+                VerboseWarnings =
+                    case proplists:get_bool(verbose, Options) of
+                        true -> check_line_lengths("nofile", PrintWidth, Formatted);
+                        false -> []
+                    end,
+                {ok, unicode:characters_to_list(Formatted), Warnings ++ VerboseWarnings};
             {skip, RawString} ->
                 {skip, RawString}
         end
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -167,7 +167,12 @@ format_range(FileName, StartLocation, EndLocation, Options) ->
             {ok, NodesInRange} ->
                 Result = format_nodes(NodesInRange, PrintWidth),
                 verify_nodes(FileName, NodesInRange, Result),
-                {ok, unicode:characters_to_binary(Result), Warnings};
+                VerboseWarnings =
+                    case proplists:get_bool(verbose, Options) of
+                        true -> check_line_lengths(FileName, PrintWidth, Result, StartLocation);
+                        false -> []
+                    end,
+                {ok, unicode:characters_to_binary(Result), Warnings ++ VerboseWarnings};
             {options, PossibleRanges} ->
                 {options, PossibleRanges}
         end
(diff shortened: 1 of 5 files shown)


michalmuskala:
Hi @ampleyfly, thank you for contributing this! I'm sorry for a delay in review, it's somewhat of a busy period for @awalterschulze and me. I hope we'll be able to get to this issue this week, so I'll ask you for couple more days of patience, if you don't mind.

test/erlfmt_SUITE.erl
```diff
@@ -1303,10 +1307,42 @@ contains_pragma_string(String) ->
...
     %% check that insert_pragma_nodes doesn't insert a pragma, when one has already been inserted.
-    ?assertEqual(
-        {ok, StringWithPragma, []},
+    ?assertMatch(
+        {ok, #{formatted := StringWithPragma, warnings := []}},
         erlfmt:format_string(StringWithPragma, [{pragma, insert}])
     ),
     StringWithPragma.
+
+overlong_warning(Config) when is_list(Config) ->
```
awalterschulze, line 1319: Really love how this is tested, thank you so much for including this.

src/erlfmt.erl
```diff
@@ -58,7 +58,13 @@ init(State) ->
 
 %% API entry point
 -spec format_file(file:name_all() | stdin, config()) ->
-    {ok, [unicode:chardata()], [error_info()]} | {skip, string()} | {error, error_info()}.
+    {ok, #{
```
awalterschulze, line 61: We were thinking it might be better to extend config with a verbose prop
```erlang
-type config() :: [{pragma, pragma()} | {print_width, pos_integer()} | verbose].
```
then we can avoid checking the line length in cases where users didn't specify the verbose flag, to speed up the common case.

If there are warnings, we think it would also simplify things to then just append them to the single `warnings` list.

What do you think?

test/erlfmt_SUITE.erl
awalterschulze: Maybe it would be good to add a unicode example, to test that we use the correct way of checking length.

src/erlfmt.erl
elbrujohalcon: Would it be too bad to keep the 3-element tuple here (i.e. `{ok, [unicode:chardata()], #{warnings => …}}`)?
When we use this function from `rebar3_format`, we just ignore the third element 🙄

elbrujohalcon (changes requested):

src/erlfmt.erl
awalterschulze: Yes we are also proposing to keep the 3 tuple, sorry if this was unclear.

test/erlfmt_SUITE.erl
ampleyfly: I guess the point would be to check that we don't get the warning, because we are counting glyphs and not bytes. That sounds like a good idea. Do you have a good example to go with?
I was thinking of using a pangram, either "色は匂へど 散りぬるを 我が世誰ぞ 常ならむ 有為の奥山 今日越えて 浅き夢見じ 酔ひもせず" or "Nne, nna, wepụ he’l’ụjọ dum n’ime ọzụzụ ụmụ, vufesi obi nye Chukwu, ṅụrịanụ, gbakọọnụ kpaa, kwee ya ka o guzoshie ike; ọ ghaghị ito, nwapụta ezi agwa." (but broken into multiple lines).
The first one is nicer to look at, but the second one is trickier.
There are examples (like this horrible thing: "A͋͠r͍̞̫̜͌ͦ̈́͐ͅt̼̭͞h́u̡̙̞̘̙̬͖͓rͬͣ̐ͮͥͨ̀͏̣") where both erlang and python3 report a length of 39, which I'm not sure can be handled.

src/erlfmt.erl
ampleyfly: Done

test/erlfmt_SUITE.erl
awalterschulze: Anyone will do, we just need at least one normal example.
I see you have added one, it looks good to me.

awalterschulze (approved):
This looks great thank you

PR #213, opened 2020-10-28
exclude files

### LINKED ISSUE #208: Ability to exclude files
erlfmt currently accepts a `files` option to pass files or globs for erlmft to check.

We'd also find it helpful to have an `excludes` option: `{excludes, ["glob1", "glob2"]}` to avoid checking certain files.

Thanks for considering this feature request, I appreciate your work!

awalterschulze:
Fixes #208 

Thanks to @TD5 for pairing with me on this

### DIFF
--- a/src/erlfmt_cli.erl
+++ b/src/erlfmt_cli.erl
@@ -254,7 +254,7 @@ read_stdin(Data) ->
         {error, Reason} -> {error, Reason}
     end.
 
--spec parse_opts(list(string())) -> parsed().
+-spec parse_opts(list()) -> parsed().
 parse_opts(Args) ->
     parse_opts(Args, [], [], #config{}).
 
--- a/src/erlfmt_cli.erl
+++ b/src/erlfmt_cli.erl
@@ -24,7 +24,7 @@
     out = standard_out :: out()
 }).
 
--type parsed() :: {format, list(), #config{}} | help | version | {error, string()}.
+-type parsed() :: {format, list(), list(), #config{}} | help | version | {error, string()}.
 
 -spec opts() -> [getopt:option_spec()].
 opts() ->
@@ -47,6 +47,9 @@ opts() ->
             "Insert a @format pragma to the top of formatted files when pragma is absent. "
             "Works well when used in tandem with --require-pragma, "
             "but it is not allowed to use require-pragma and insert-pragma at the same time."},
+        {exclude_files, undefined, "exclude-files", string,
+            "files not to format. "
+            "This overrides the files specified to format"},
         {files, undefined, undefined, string,
             "files to format, - for stdin. "
             "If no files are provided and one option of [-w, -o, -c] are provided then "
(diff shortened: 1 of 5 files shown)


src/erlfmt_cli.erl
```diff
@@ -92,39 +96,45 @@ with_parsed(Name, Config) ->
             erlang:halt(127)
     end.
 
+set_difference(Files, Excludes) ->
+    sets:to_list(sets:subtract(sets:from_list(Files), sets:from_list(Excludes))).
```
michalmuskala, line 100: I think in this case list subtraction could also work `Files -- Excludes`. Is there a specific reason for going with the set approach?

src/erlfmt_cli.erl
```diff
@@ -246,48 +256,54 @@ read_stdin(Data) ->
...
-parse_opts([], Files, Config) ->
+parse_opts([], [stdin], [], Config) ->
+    {format, [stdin], [], Config};
+parse_opts([], [stdin], [_Exclude | _], _Config) ->
+    {error, "stdin mode can't be combined with excluded files"};
+parse_opts([], Files, Exclude, Config) ->
     case lists:member(stdin, Files) of
         true -> {error, "stdin mode can't be combined with other files"};
-        false -> {format, lists:reverse(Files), Config}
+        false -> {format, lists:reverse(Files), lists:reverse(Exclude), Config}
```
michalmuskala, line 304: What if we checked files vs excludes in here and otherwise produced an `{error, ...}` tuple like we do for other cases? I think this could simplify `unprotected_with_config` that is already rather complex

src/erlfmt_cli.erl
awalterschulze: list difference would only work if the lists contain no duplicates.
```erlang
[{files, [a.erl, a.erl]}, {files, [a.erl]}, {exclude_files, [a.erl]}]
```

src/erlfmt_cli.erl
```diff
@@ -305,8 +321,10 @@ resolve_parsed(PreferParsed, DefaultParsed) ->
             version;
         {_, version} ->
             version;
-        {{format, PreferFiles, PreferConfig}, {format, DefaultFiles, DefaultConfig}} ->
+        {{format, PreferFiles, PreferExclude, PreferConfig},
+            {format, DefaultFiles, DefaultExclude, DefaultConfig}} ->
             {format, resolve_files(PreferFiles, DefaultFiles),
+                resolve_files(PreferExclude, DefaultExclude),
                 resolve_config(PreferConfig, DefaultConfig)}
```
michalmuskala, line 328: Is this correct? Should one excludes apply to the file list from the other source?

src/erlfmt_cli.erl
awalterschulze: This was the first thing we considered, but what about specifying no files and only excluding files.
This should combine in a logical way with the default files we provide or those supplied in the rebar config.

src/erlfmt_cli.erl
awalterschulze: Yes I think this is correct, but maybe you have an example we haven't thought of

michalmuskala (approved):

PR #214, opened 2020-10-30
analysis of comments

awalterschulze:
- We move cloning inaka into a script that can be referenced rather than copied into every single readme
  - We did an analysis of comments in practice

You can view the analysis [here](https://github.com/WhatsApp/erlfmt/blob/dfe84e232b0309e94e1f4a1c2f35046c6da11c6f/doc/FormattingDecisionComments/Readme.md), since reading it in the diff won't be as fun.


elbrujohalcon (approved):

doc/FormattingDecisionComments/columns.html
```diff
@@ -0,0 +1,86 @@
...
+
+
+
+Plotly.newPlot('myPlot', data, layout);
+
+</script>
+
+
+</body>
+</html>
```
michalmuskala, line 85: the html plots are a nice touch, but it's rather hard to see them from github. Should we worry about that?

awalterschulze: I tried gnuplot and matplotlib but gave up.
Both had issues installing or crashed while running, so I decided to limit the time that I spent on this.
It is inconvenient to view on github though :(
You have to download the raw file to view, I am sorry.
I hope it is okay this way though.

michalmuskala (approved):

PR #232, opened 2020-12-07
Print literal value instead of crashing on missing text annotation

richcarl:
This allows you to format syntax trees that have been synthesized outside erlfmt without forcing the caller to add "text" annotations for all literals in order to print them.

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -115,12 +115,22 @@ expr_to_algebra(Expr) when is_tuple(Expr) ->
 expr_to_algebra(Other) ->
     do_expr_to_algebra(Other).
 
-do_expr_to_algebra({string, Meta, _Value}) ->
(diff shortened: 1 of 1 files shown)


awalterschulze:
Would it be possible to parse the AST and transform it to include these "text" annotations, before passing it back to erlfmt?

Or include a test with an explanation comment, so that we do not simplify this code later by accident?

richcarl:
It is of course possible, just a bit annoying and expensive to do an extra pass rewriting the whole AST for this.

awalterschulze:
If it is too much effort to have an extra pass over the AST, consider adding a test, with a comment that explains why this is necessary, to make sure we do not regress and remove this change when thinking we are cleaning things up.

awalterschulze:
This is probably go to merge, just need a test.
Do you want to add a test or maybe we can add one for you?

PR #235, opened 2021-01-06
Only write to file the formatted code when it is different to the current one

### LINKED ISSUE #231: erlfmt -w bumps mtime of all files it checks
My employer uses a custom build system for Erlang which can be very fast for incremental builds because it only recompiles files that have changed. However, I've noticed that the incremental aspect of it got broken once we started using `erlfmt` and periodically ran `erlfmt -w` because it modifies (presumably overwrites) any files that are checked.

Is there some fundamental reason why `erlfmt` writes to files that don't require any formatting changes? If not, I'd be happy to have a go at making a patch for this myself. In that case please let me know if there's some particular approach that the maintainers would prefer me to take here.

gonzalobf:
Fixes #231. I couldn't find any test that checks the writing to but let me know if you think it is needed and I'll try to implement one.

I used `stats` to check if it works: 
```bash
$ stat rebar.config | grep -F 'Modify'                                                                                                                                            0s
Modify: 2021-01-06 09:38:08.204419183 +0000
$ _build/release/bin/erlfmt -w "{src,include,test}/*.{hrl,erl}" rebar.config
$ stat rebar.config | grep -F 'Modify'                                                                                                                                            0s
Modify: 2021-01-06 09:38:08.204419183 +0000
```

### DIFF
--- a/src/erlfmt_cli.erl
+++ b/src/erlfmt_cli.erl
@@ -202,7 +202,14 @@ write_formatted(FileName, Formatted, Out) ->
             print_error_info({OutFileName, 0, file, Reason1}),
             error
     end,
-    case file:write_file(OutFileName, unicode:characters_to_binary(Formatted)) of
+    {ok, OriginalBin} = file:read_file(FileName),
+    case unicode:characters_to_binary(Formatted) of
+        OriginalBin -> ok;
+        FormattedBin -> write_file(OutFileName, FormattedBin)
+    end.
+
+write_file(OutFileName, FormattedBin) ->
+    case file:write_file(OutFileName, unicode:characters_to_binary(FormattedBin)) of
         ok ->
             ok;
         {error, Reason2} ->


kvelicka:
@michalmuskala any thoughts on this?

michalmuskala:
Hi @kvelicka. Thank you so much for your contribution. I'm sorry about the delay.

We've been quite busy the last couple days. I'll get back to you as soon as I can.

kvelicka:
Thanks for the quick reply! No worries and no rush, was just making sure that this was on someone's radar.

Credit for the code is all to @gonzalobf too!

awalterschulze:
We have decided to merge this as is and created an issue to tackle an optimization https://github.com/WhatsApp/erlfmt/issues/238

awalterschulze:
Thank you for bringing this to our attention and fixing it <3

PR #237, opened 2021-01-14
add library module for converting between erlfmt and syntax tools

richcarl:
Allows roundtrip conversion from erlfmt:read_nodes() to syntax tools (or erl_parse) and back to erlfmt:format_nodes().


awalterschulze:
This certainly looks useful for writing refactoring tools in erlang itself.

We are having a hard time deciding if this belongs in erlfmt or a separate library:
If we merge it in erlfmt, we would need to maintain it and if we don't merge it, then we would need to document our abstract_forms types better. To merge in erlfmt in would need a barrage of tests, that would make us confident that we can maintain it.  Are you willing to do this work?

I only skimmed the code, but I saw a lot of TODOs are they intended to be fixed before merge or ...?

richcarl:
No, the TODOs are meant for doing later, they're not blocking anything. But this code needs to find a home first, and I'm intending to do the work on it. Also hoping to work on reducing the internal differences between erlfmt and syntax tools so the conversion becomes less complicated in places.

awalterschulze:
Do you also intend to add a barrage of tests, something in the order of https://github.com/WhatsApp/erlfmt/blob/master/test/erlfmt_format_SUITE.erl ?

Knowing this would help us to make a decision.

richcarl:
That would probably be relatively easy to do, just adapting the existing tests and checking that the round-tripped code looks the same, so yes.

awalterschulze:
That would be one way to get quite a large of coverage, but feel free to take some liberty on what you think is sensible.

Do you also maybe have thoughts on why you think this is the right home for this code?

richcarl:
I think this is the right home, because the erlfmt format doesn't officially exist outside this app. By having the conversion functions here, nobody should need to know about it or depend on it. Also, it becomes easy to make synchronized changes in erlfmt internals and the conversion functions with a single commit.

awalterschulze:
We have talked and you make a compelling argument :D
Lets consider this the future home of this code, but lets also get some tests before merging.

richcarl:
I think this is ready now.

test/erlfmt_SUITE.erl
```diff
@@ -99,6 +99,13 @@ end_per_group(_GroupName, _Config) ->
     ok.
 
 init_per_testcase(_TestCase, Config) ->
+    GroupProps = ?config(tc_group_properties, Config),
+    case proplists:get_bool(syntax_tools, GroupProps) of
+        true ->
+            put('$syntax_tools$', true);
```
awalterschulze, line 105: Wow I didn't know about this magic.
I know this is a minimal diff, but thinking of the next person reading this, maybe it would be better to use a parameter, since each test already has access to Config.  Something like:

```
is_syntax_tools(Config) ->
    proplists:get_bool(syntax_tools, ?config(tc_group_properties, Config)).
    
 some_test(Config) ->
   RoundTripWithSyntaxTools = is_syntax_tools(Config)
   ...
   parse_form(Form, RoundTripWithSyntaxTools)
```

But I think this was a great way to get test coverage and it is really impressive that everything is passing.

src/erlfmt_ast.erl
```diff
@@ -0,0 +1,1086 @@
...
+            st_to_erlfmt(Guard)
+    end.
+
+st_typedecl_to_erlfmt({atom, _, type} = Tag, [Term], Pos) ->
+    %% the argument is raw erl_parse data here, lifted to abstract form by
+    %% erl_syntax, so it first needs to be made concrete again
+    {Name, Type, Args} = erl_syntax:concrete(Term),
+    Name1 = {atom, dummy_anno(Name), Name},
+    Type1 = st_to_erlfmt(Type),
+    %exit({Name1, Type, Type1}),
```
awalterschulze, line 1024: What is this comment for?

src/erlfmt_ast.erl
```diff
@@ -0,0 +1,1086 @@
...
+            CAnno = dummy_anno(),
+            {'try', Pos, Body, {clauses, CAnno, Clauses}, {clauses, CAnno, Handlers}, After};
+        comment ->
+            {comment, Pos, erl_syntax:comment_text(Node)};
+        _ ->
+            st_to_erlfmt_1(Node)
+    end.
+
+st_to_erlfmt_1(Node) ->
+    %% TODO: should we convert full erl_syntax pos+annotation to erlfmt anno?
```
awalterschulze, line 860: It seems like you are doing Anno conversions here, so maybe this TODO is already done or I don't understand it, sorry?

src/erlfmt_ast.erl
```diff
@@ -0,0 +1,1086 @@
...
+erlfmt_clause_to_st(Pos, Patterns, Guard, Body) ->
+    Groups = [
+        Patterns,
+        [erlfmt_guard_to_st(Guard)],
+        [erlfmt_to_st(B) || B <- Body]
+    ],
+    erl_syntax:set_pos(erl_syntax:make_tree(clause, Groups), Pos).
+
+%% New `{guard_or, Anno, GuardAndList}` and `{guard_and, Anno, Exprs}` nodes
+%% are introduced to support annotating guard sequences, insted of a plain
```
awalterschulze, line 501: instead instead of insted ;)

src/erlfmt_ast.erl
```diff
@@ -0,0 +1,1086 @@
...
+        [
+            erlfmt_to_st(Subtree)
+            || Subtree <- Group
+        ]
+        || Group <- Groups
+    ].
+
+get_function_name([{clause, _, {call, _, Name, _}, _, _} | _]) ->
+    %% take the name node of the first clause with a call shape
+    %% TODO: this loses info if not all clauses have the same name
```
awalterschulze, line 450: So is this future work and do we have a missing test case?

awalterschulze (commented):
Overall have to say it looks amazing.
The comments are mostly nitpicks and I am wondering what the strategy is for the TODOs.
Some of them seem related to syntaxtools and others to erlfmt.
Should we open issues to tackle them or what do you think?

test/erlfmt_SUITE.erl
richcarl: I like to pass things as parameters whenever I can, but all the test code would have to get the extra Options parameter passed all the way down to the parse_form/parse_forms functions, and that seemed too intrusive. I also thought about using meck to modify the behaviour of the test code, but that's also a bit yucky and causes a dep on meck. What do you prefer?

src/erlfmt_ast.erl
richcarl: Because I'm daft and forgot to remove it. It was there to provoke a visible crash at some point while debugging.

src/erlfmt_ast.erl
richcarl: erl_syntax nodes have a separate annotation field (because the old annotation field used to be only for line numbers), and that could contain arbitrary "user data", so the question is whether to pick up any such annotations and inject them into the erlfmt annotation when going from st to erlfmt, or just abandon them (as done now).

src/erlfmt_ast.erl
richcarl: There is future work to maybe extend erl_syntax to preserve separate name nodes for each clause. Right now, nothing can be done here.

src/erlfmt_ast.erl
richcarl: fixed

src/erlfmt_ast.erl
richcarl: fixed

richcarl:
I'm currently pondering what changes I can make in syntax tools that will preserve compatibility with existing code that's using it, but being better at representing the same things under the hood that erlfmt does. Some of those TODOs require that sort of change first. Some other things may be possible to just change on the erlfmt side (after consulting you) and for those I could open tickets if you like.

src/erlfmt_ast.erl
awalterschulze: Aha, well I can say that only the relative line numbers are taken into account by the formatting algorithm.
So then the question would be, what would be the use case to preserve them and is that worth the effort.

test/erlfmt_SUITE.erl
awalterschulze: I think I prefer a parameter, since I don't think it needs to pass too far down.
But let's also hear what @michalmuskala has to say.

awalterschulze:
Is it possible to label the TODOs in the code as syntax_tools TODOs and actual TODOs for this code, to try and make it clear what is do-able by someone only looking at this code?  It would at least help me to discern between them.

test/erlfmt_SUITE.erl
michalmuskala: I think this is fine. It's not pretty, but it's practical. I actually wonder if we should try doing something in `erlfmt_format_SUITE` as well to leverage all the examples we have in there, though I'm not 100% sure how that would look like.

maxno-kivra:
What happened to this? I wanted to write an `inspect` function like Elixir's today, and saw that there's such an algebra here in erlfmt. Ehere's no convenient way to convert a term to the internal AST, like `erl_syntax:abstract/1`, and I can't use that one and then convert from `erl_syntax` to `erlfmt`.

Also, when debugging macros and parse transforms, it's nice to be able to pretty print AST, which again would have been trivial if this was merged. 🙏

richcarl:
> What happened to this?

It got left by the road when I moved on to other things. Thanks for reminding me it still exists - it would be nice to get it done. It might have become a bit stale in the meantime though.

Issue #125, opened 2020-09-02
Comments: insert_expr_list is sometimes used incorrectly and moves some comments


michalmuskala:
```erlang
%% from https://github.com/erlang/otp/blob/66b28bdd51899b60e0bb44d46ae4d0ae6b89601f/lib/compiler/test/trycatch_SUITE.erl#L1107-L1118
grab_bag_3() ->
    try 2 of
        true ->
            <<
              "" || [V0] = door
            >>
    catch
        error:true:V0 ->
            []
            %% The default clause here (which re-throws the exception)
            %% would not return two values as expected.
    end =:= (V0 = 42).
```

```
=ERROR REPORT==== 2-Sep-2020::14:40:35.075619 ===
Error in process <0.70.0> with exit value:
{unsupported,
    [{erlfmt_format,do_expr_to_algebra,
         [{comment,
              #{end_location => {11,56},location => {10,13}},
              ["%% The default clause here (which re-throws the exception)",
               "%% would not return two values as expected."]}],
         [{file,"/Users/micmus/github/WhatsApp/erlfmt/src/erlfmt_format.erl"},
          {line,224}]},
     {erlfmt_format,expr_to_algebra,1,
         [{file,"/Users/micmus/github/WhatsApp/erlfmt/src/erlfmt_format.erl"},
          {line,110}]},
     {erlfmt_format,try_after_to_algebra,1,
         [{file,"/Users/micmus/github/WhatsApp/erlfmt/src/erlfmt_format.erl"},
          {line,705}]},
     {erlfmt_format,try_to_algebra,4,
         [{file,"/Users/micmus/github/WhatsApp/erlfmt/src/erlfmt_format.erl"},
          {line,697}]},
     {erlfmt_format,expr_to_algebra,1,
         [{file,"/Users/micmus/github/WhatsApp/erlfmt/src/erlerlfmt
```

michalmuskala:
Command I used to find this:

```
_build/release/bin/erlfmt ~/github/erlang/otp/lib/*/{src,include,test}/*.{e,h}rl > /dev/null
```

awalterschulze:
Looking into this there is another problem
Try to format this:
```erlang
try 2 of
    true -> ok
catch
    _ ->
        undefined
        %% after catch
after
    ok
    %% after after
end
```
We get:
```erlang
try 2 of
    true -> ok
catch
    _ ->
        undefined
after
    %% after catch

    ok
    %% after after
end
```

[...]

PR #242, opened 2021-01-21
Adjust AST for try clauses such that they do not move comments around

awalterschulze:
Tackled parts of https://github.com/WhatsApp/erlfmt/issues/125 , but issues with receive statements persist and will be tackled at a later date.

### DIFF
--- a/test/erlfmt_format_SUITE.erl
+++ b/test/erlfmt_format_SUITE.erl
@@ -1711,13 +1711,13 @@ receive_expression(Config) when is_list(Config) ->
         "    2 -> two\n"
         "end\n"
     ),
-    ?assertSame(
-        "receive\n"
-        "after\n"
-        "    % foo\n"
-        "    0 -> ok\n"
-        "end\n"
-    ),
+    % ?assertSame(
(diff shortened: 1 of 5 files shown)


test/erlfmt_format_SUITE.erl
```diff
@@ -1710,6 +1710,30 @@ receive_expression(Config) when is_list(Config) ->
...
+        "    %% after after for receive\n"
+        "end\n",
+        "receive\n"
+        "    1 -> ok\n"
+        "%% after receive\n"
+        "\n"
+        "after 0 ->\n"
+        "    ok\n"
+        "    %% after after for receive\n"
+        "end\n"
```
michalmuskala, line 1736: Let's maybe keep the receive changes completely out of this. What do you think?

michalmuskala (approved):

test/erlfmt_format_SUITE.erl
awalterschulze: I think it is easy to modify when we change the behaviour, rather than rewrite the test again once we start working on this again.

PR #254, opened 2021-02-22
Support macros in type names

### LINKED ISSUE #253: Parser can't handle macro in module-qualifier within types
The below examples don't parse and it feels like it would be easy to fix (but there might be some reason why macros are not allowed in these positions)
```
-spec ?MODULE:f() -> t().
-spec f() -> ?MODULE:t().
```
```
-spec mod:?FOO() -> t().
-spec f() -> mod:?FOO().
```

michalmuskala:
Closes #253


awalterschulze (approved):

awalterschulze:
Does this solve https://github.com/WhatsApp/erlfmt/issues/253 ?

michalmuskala:
It only did partially before, but now all cases from the issue should be handled

awalterschulze:
Very nice thank you :D

PR #255, opened 2021-03-03
more options considered for placement of when for multiline guards

awalterschulze:


michalmuskala (approved):

doc/FormattingDecisionWhenMultilineGuards.md
```diff
@@ -204,6 +213,83 @@ bar(X) when
 -> body.
 ```
 
+## When on same line as guards
+
+We can move `when` to the same line as guards and still keep the `->` on a separate line.
+This creates an even greater differentiation between guards and body, than we only the `->` is on a separate line.
```
elbrujohalcon, line 219: we -> when

(I think)

awalterschulze: Thank you great catch :)

PR #259, opened 2021-03-17
Break tuples with container in first position

### LINKED ISSUE #153: Tuples: Deeply nested tuples get formatted strangely
```
{{<<"key3">>}, {[{<<"true">>, true}, {<<"false">>, false}, {<<"nullqweqwewqdqwqdwqdq">>, null}, {<<"numberqweqwrqwdqd">>, 1}]}}.
```
This still formats ok (note the correct alignment of opening and closing parens):
```
{{<<"key3">>},
    {[
        {<<"true">>, true},
        {<<"false">>, false},
        {<<"nullqweqwewqdqwqdwqdq">>, null},
        {<<"numberqweqwrqwdqd">>, 1}
    ]}}.
```
But wrapping the second top level element into another tuple leads to strange results:
```
{{<<"key3">>},
    {{[
            {<<"true">>, true},
            {<<"false">>, false},
            {<<"nullqweqwewqdqwqdwqdq">>, null},
            {<<"numberqweqwrqwdqd">>, 1}
        ]}}}.
```
But perhaps it's intended. I'm not really sure here.

michalmuskala:
This is still work in progress.

This changes how tuples are laid out when the first element is a complex term - they are printed like lists and other containers. The main use-case for the "flex" tuple layout we used before was to support nicely printing tagged tuples - this is not changed.

Fixes #153


awalterschulze:
- Tuples inside tuples are no tightly coupled `{{A}}` with no_break in between, which means tuples inside tuples do not have a weird indentation, see the `is_container` function
  - When the first element in a tuple is not a tag, the tuple is now formatted as if it was a list, see the `is_tag` function.

awalterschulze (approved):

Issue #260, opened 2021-03-20
Parser crashes on bad record


gomoripeti:
The parser crashes on a bad record while it wants to emit a meaningful error

```
1> erlfmt:read_nodes_string("nofile","-record(r, {12}).").
** exception error: bad argument
     in function  erl_anno:anno_info/1
        called as erl_anno:anno_info(#{end_location => {1,15},location => {1,13},text => "12"})
     in call from erlfmt_parse:ret_err/2 (/Users/ext.pgomori/git/erlfmt/src/erlfmt_parse.yrl, line 1059)
     in call from erlfmt_parse:record_tuple/1 (/Users/ext.pgomori/git/erlfmt/src/erlfmt_parse.yrl, line 1042)
     in call from erlfmt_parse:build_attribute/3 (/Users/ext.pgomori/git/erlfmt/src/erlfmt_parse.yrl, line 1035)
```

I think there is a mixup (maybe just leftover from `erl_parse`) between `erl_anno:anno()` and `erlfmt_scan:anno()`
Probably this should also be `erlfmt_scan:token()` https://github.com/WhatsApp/erlfmt/blob/master/src/erlfmt_parse.yrl#L967

Additionally `erlfmt_parse:record_tuple/1` seems to prevent parsing a macro as a record field, eg
```
-define(F1, field1).
-record(r, {?F1}).
```

awalterschulze:
Thank you for reporting.  We will look into it asap.

PR #261, opened 2021-03-24
fix parsing of bad records

awalterschulze:
Fixes https://github.com/WhatsApp/erlfmt/issues/260

Seems `erl_anno:location` crashes, since we have a `erlfmt_scan:anno()` and not a type from `erl_anno`.
I changed to it rather call `erlfmt_scan:get_anno(location, Anno)` as I assume this is what was intended.

Also fixed the type annotation for `token()` to point to `erlfmt_scan:token()` as pointed out by @gomoripeti 
```
-spec parse_node(Tokens) -> {ok, AbsNode} | {error, ErrorInfo} when
      Tokens :: [token()],
      AbsNode :: abstract_node(),
      ErrorInfo :: error_info()
```

Since it seems the tokens passed to parse_node come from

```
-spec string_node(string()) -> node_ret().
```

where `node_ret()` is:

```
-type node_ret() ::
    {ok, [token()], [comment()], state()}
    | {error, {erl_anno:location(), module(), term()}, erl_anno:location()}
    | {eof, erl_anno:location()}.
```

and here `token()` refers to `erlfmt_scan:token()`

### DIFF
--- a/test/erlfmt_SUITE.erl
+++ b/test/erlfmt_SUITE.erl
@@ -1386,18 +1386,42 @@ overlong_warning(Config) when is_list(Config) ->
     ?assertEqual([], [Line || {8, _} = Line <- RangeLongLines]).
 
 do_not_crash_on_bad_record(Config) when is_list(Config) ->
+    %% normal record
+    ?assertMatch(
+        {ok, [{attribute, _, {atom, _, record}, _}], _},
+        erlfmt:read_nodes_string("nofile", "-record(r, { field :: string() }).")
+    ),
     %% bad record declaration
-    {ok, _, _} = erlfmt:read_nodes_string("nofile", "-record(r, [1])."),
+    {ok, [{exprs, _, _}], _} = erlfmt:read_nodes_string("nofile", "-record(r, [1])."),
     %% bad record field
-    {ok, _, _} = erlfmt:read_nodes_string("nofile", "-record(r, {2})."),
+    {ok, [{exprs, _, _}], _} = erlfmt:read_nodes_string("nofile", "-record(r, {2})."),
     %% bad record macro field
-    {ok, _, _} = erlfmt:read_nodes_string(
+    {ok, [_, Attr], _} = erlfmt:read_nodes_string(
         "nofile",
         "-define(F1, field1).\n"
         "-record(r, {?F1}).\n"
     ),
+    ?assertMatch(
+        {attribute, _, {atom, _, record}, [
+            {atom, _, r},
+            {tuple, _, [
+                {record_field, _, {macro_call, _, _, _}}
+            ]}
+        ]},
+        Attr
+    ),
     %% bad record macro field 2
-    {ok, _, _} = erlfmt:read_nodes_string(
+    {ok, [Attr2], _} = erlfmt:read_nodes_string(
         "nofile",
         "-record(foo, {?FOO, b :: any()})."
+    ),
+    ?assertMatch(
+        {attribute, _, {atom, _, record}, [
+            {atom, _, foo},
+            {tuple, _, [
+                {record_field, _, {macro_call, _, _, _}},
+                {op, _, _, {record_field, _, _}, {call, _, _, _}}
+            ]}
+        ]},
+        Attr2
     ).
--- a/test/erlfmt_SUITE.erl
+++ b/test/erlfmt_SUITE.erl
@@ -80,7 +80,8 @@
     broken_range/1,
     contains_pragma/1,
     insert_pragma/1,
-    overlong_warning/1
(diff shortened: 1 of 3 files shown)


test/erlfmt_SUITE.erl
```diff
@@ -1382,3 +1384,15 @@ overlong_warning(Config) when is_list(Config) ->
     ?assertEqual([], [Line || {8, _} = Line <- StringLongLines]),
     ?assert(lists:member({6, 121}, RangeLongLines)),
     ?assertEqual([], [Line || {8, _} = Line <- RangeLongLines]).
+
+do_not_crash_on_bad_record(Config) when is_list(Config) ->
+    %% bad record declaration
+    {ok, _, _} = erlfmt:read_nodes_string("nofile", "-record(r, [1])."),
+    %% bad record field
+    {ok, _, _} = erlfmt:read_nodes_string("nofile", "-record(r, {2})."),
+    %% bad record macro field
```
gomoripeti, line 1393: this is actually valid code and the erlfmt_parse grammar allows macros in this position https://github.com/WhatsApp/erlfmt/blob/ec86fb280de822c3989bb14b28e7879f79532f7d/src/erlfmt_parse.yrl#L373

maybe something like this change would allow it to go through
```
- record_fields([{op,Am,'=',{atom,Aa,A},Expr}|Fields]) ->
-     [{record_field,Am,{atom,Aa,A},Expr}|record_fields(Fields)];
+ record_fields([{op,Am,'=',FieldName,Expr}|Fields]) ->
+    [{record_field,Am,FieldName,Expr}|record_fields(Fields)];
```

gomoripeti (commented):
🙏

test/erlfmt_SUITE.erl
awalterschulze: I changed it back to `erl_anno:location` in `ret_err` and then tried this change, but it still broke.

The current pull request still returns the parse record as:
```
[{attribute,
      #{end_location => {1,21},location => {1,1}},
      {atom,
          #{end_location => {1,8},location => {1,2},text => "define"},
          define},
      [{var,#{end_location => {1,11},location => {1,9},text => "F1"},'F1'},
       {atom,
           #{end_location => {1,19},location => {1,13},text => "field1"},
           field1}]},
  {exprs,
      #{dot => true,end_location => {2,19},location => {2,1}},
      [{op,#{end_location => {2,18},location => {2,1}},
           '-',
           {call,
               #{end_location => {2,18},location => {2,2}},
               {atom,
                   #{end_location => {2,8},location => {2,2},text => "record"},
                   record},
               [{atom,
                    #{end_location => {2,10},location => {2,9},text => "r"},
                    r},
                {tuple,
                    #{end_location => {2,17},location => {2,12}},
                    [{macro_call,
                         #{end_location => {2,16},location => {2,13}},
                         {var,
                             #{end_location => {2,16},
                               location => {2,14},
                               text => "F1"},
                             'F1'},
                         none}]}]}}]}]
                         ```
                         
which I think is good enough
Parsing it as a `record_field` might presume too much, I mean technically it could be defining two record fields.
We do not necessarily want to format it was `record_field` in this case formatting it as a `macro_call` should be fine.

michalmuskala: I'm afraid that parsing as `exprs` will fail when using more specific record syntax. In particular something like `-record(foo, {?FOO, b :: any()})`.

I think parsing it as `record_field` with `?FOO` as the name makes sense. Yes, it's possible there's more expressions hidden there, but that shouldn't impact how we parse it, as long as it looks like a single expression.

gomoripeti: sorry, my diff suggestion was incorrect/incomplete, and I also didn't know that the parser would return an `exprs` element (I assumed it just gives up and returns an error)
it's up to you how you handle these scenarios

awalterschulze: I updated it with @gomoripeti 's suggestion

awalterschulze: Ah sorry, I see we want some of these bad records to actually parse as `{attribute ...` instead of `{exprs....`

awalterschulze:
The current solution breaks when you try to format a record with a macro.  Investigating, but if I have to guess it has to do with `record_field` not having an `op`

awalterschulze:
> The current solution breaks when you try to format a record with a macro. Investigating, but if I have to guess it has to do with `record_field` not having an `op`

wrong guess, I didn't capture the annotation properly in the record_field

awalterschulze:
All is ready to review again.
Sorry for misunderstanding before.

michalmuskala (approved):
Looks great!

Issue #163, opened 2020-09-19
Binary Operators: Strange formatting of case + pattern match


david-kubecka:
```
A = {B, C} = case X of true -> f(); _ -> g() end.
```
gets formatted into
```
A =
    {B, C} =
    case X of
        true -> f();
        _ -> g()
    end.
```
I would expect either
```
A =
    {B, C} =
        case X of
            true -> f();
            _ -> g()
        end.
```
or preferably
```
A = {B, C} =
    case X of
        true -> f();
        _ -> g()
    end.
```

awalterschulze:
This is a very interesting case.  I don't think we have considered this before. Thank you for bringing this up.

michalmuskala:
Interestingly the Elixir formatter does exactly the same thing.
```elixir
a = {b, c} = case d do e -> e end
```
gets formatted to
```elixir
a =
  {b, c} =
  case d do
    e -> e
  end
```

michalmuskala:
I'm going to copy over the reasoning for this behaviour in Elixir that José provided, I find it convincing myself.

> In my opinion, it is intentional. What I do in this case is to replace the operators and expressions and ask if it looks natural. For example, I do expect this:
>
> ```
> foo ++
>  bar ++
>  baz
> ```
> I think indenting baz another time would feel unnatural. So I don’t think we should do it for case. There is an argument though that the first two should be on the same line but it is relatively hard to do with right precedence operators.

awalterschulze:
A great explanation.  It sounds like we would have to make a big exception for equal.

I guess I could also write:
```erlang
A = {{B1, B2}, C} = {B, C} =
    case X of
        true -> f();
        _ -> g()
    end.
```

I also expect this to be very tricky.
@michalmuskala What challenges do you expect with right associativity?

michalmuskala:
The right-precedence makes it complex, because of how expressions are grouped and represented in the AST. This should be more visible if we add explicit parenthesis to represent the grouping:

```erl
A = ({{B1, B2}, C} = ({B, C} =
    case X of
        true -> f();
        _ -> g()
    end)).
```

With right-associative operators we lay out this nested expression
```erl
{B, C} =
    case X of
        true -> f();
        _ -> g()
    end
```
before laying out the wrapping expressions.

We could potentially do a rotation on the AST tree in case the meaning of the resulting program would be the same (i.e. in places where `=` being left- or right-associative would produce the same program). But it is somewhat of a tricky operation.

[...]

Issue #161, opened 2020-09-19
Binary Operators: Wrong alignment of boolean operators


david-kubecka:
Following is IMO well formatted. Note the extra alignment of the second `orelse` argument. That is desired because the previous operator is different.
```
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa andalso
    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb orelse
        cccccccccccccccccccccccccccccccccc.
```
In reality this gets formatted into
```
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa andalso
    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb orelse
    cccccccccccccccccccccccccccccccccc.
```
which somehow obscures the logic.

awalterschulze:
Very interesting.
What is the operator precedence here I assume `andalso` is bound stronger than `orelse` as in:
`(a andalso b) orelse c`

michalmuskala:
This looks like a bug to me. I believe we should be always increasing the indentation after an operator, unless it's a repeated sequence of the same operator. The rule is to increase, though, not increasing is an exception.

awalterschulze:
Sounds good to me :)

awalterschulze:
Looking at this again I don't know if this really is a bug.
The following tests both pass:

```erlang
?assertSame(
        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA orelse\n"
        "    BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB andalso\n"
        "        CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC.\n"
    ),
    ?assertSame(
        "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa andalso\n"
        "    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb orelse\n"
        "    cccccccccccccccccccccccccccccccccc.\n"
    ),
```

The lowercase parses as:
`{orelse, {andalso, aaaa, bbbb}, cccc}`
and the capital letter parses as:
`{orelse, AAAA, {andalso, BBBB, CCCC}`

It is clear that the capital letters need to indent, since the `andalso` is a right child of the `orelse`.
In the lowercase letter case the `aaaa` and `bbbb` are technically "equal" to the `cccc`, so having them on the same indentation level makes sense.

[...]

PR #262, opened 2021-03-24
rewrite binop using associativity

### LINKED ISSUE #136: Binary Operators: Last binary operator not formatted consistently

awalterschulze:
This replaces the work done in https://github.com/WhatsApp/erlfmt/pull/198 and Fixes #136 and #149 and https://github.com/WhatsApp/erlfmt/issues/163

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -316,10 +316,18 @@ binary_op_to_algebra(Op, Meta0, Left0, Right0) ->
 rewrite_assoc(Op, MetaA, A, {op, MetaBC, Op, B, C} = BC) ->
     case erlfmt_scan:get_anno(parens, MetaBC, false) of
         true -> {op, MetaA, Op, A, BC};
-        _ -> rewrite_assoc(Op, MetaA, rewrite_assoc(Op, MetaBC, A, B), C)
+        _ ->
+            AB = rewrite_assoc(Op, update_meta_location(MetaBC, A, B), A, B),
+            rewrite_assoc(Op, update_meta_location(MetaA, AB, C), AB, C)
     end;
 rewrite_assoc(Op, Meta, Left, Right) ->
     {op, Meta, Op, Left, Right}.
+
+update_meta_location(Meta, First, Last) ->
+    Meta#{
+        location := erlfmt_scan:get_anno(location, First),
+        end_location := erlfmt_scan:get_anno(end_location, Last)
+    }.
 
 binary_op_to_algebra(Op, Meta, Left, Right, Indent) ->
     LeftD = expr_to_algebra(Left),
@@ -360,33 +368,9 @@ is_call({macro_call, _, _, _}) -> true;
 is_call(_) -> false.
 
 breakable_binary_op_to_algebra(OpD, Left, Right, LeftD, RightD, Indent) ->
-    HasBreak = has_break_between_op(OpD, Left, Right),
+    HasBreak = has_break_between(Left, Right),
     RightOpD = nest(break(concat(<<" ">>, OpD), group(RightD)), Indent, break),
     concat(group(LeftD), group(concat(maybe_force_breaks(HasBreak), RightOpD))).
-
-%% has_break_between_op has to do a special check for a break between binary operators,
-%% since by this time the binary operator's in the AST has been rewritten:
-%% from `(1 + (2 + (3 + 4)))` to `(((1 + 2) + 3) + 4)`
-%% which means when we check for a break in
-%% (((1 +
-%%    2) + 3) + 4)
-%% when we get the left as
-%% ((1 +
-%%    2) + 3)
-%% and the right is 4
-%% then we want to check for a break between 3 and 4 not between 1 and 4
-has_break_between_op(OpD, {op, Meta, Op, _A, B} = AB, C) ->
-    case OpD == string(atom_to_binary(Op, utf8)) of
-        true ->
-            case erlfmt_scan:get_anno(parens, Meta, false) of
-                false -> has_break_between_op(OpD, B, C);
-                _ -> has_break_between(AB, C)
-            end;
-        false ->
-            has_break_between(AB, C)
-    end;
-has_break_between_op(_OpD, Left, Right) ->
-    has_break_between(Left, Right).
 
 union_to_algebra(Meta, Left, Right) ->
     Doc = break(expr_to_algebra(Left), fold_unions(Right)),
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -307,10 +307,19 @@ binary_op_to_algebra('/', _Meta, {atom, _, _} = Left, {integer, _, _} = Right) -
 binary_op_to_algebra('|', Meta, Left, Right) ->
     %% | does not increase indentation and breaks as a whole
     union_to_algebra(Meta, Left, Right);
-binary_op_to_algebra(Op, Meta0, Left, Right) ->
+binary_op_to_algebra(Op, Meta0, Left0, Right0) ->
     %% don't print parens and comments twice - expr_to_algebra took care of it already
     Meta = erlfmt_scan:delete_annos([parens, pre_comments, post_comments], Meta0),
+    {op, _, _Meta, Left, Right} = rewrite_assoc(Op, Meta, Left0, Right0),
     binary_op_to_algebra(Op, Meta, Left, Right, ?INDENT).
+
+rewrite_assoc(Op, MetaA, A, {op, MetaBC, Op, B, C} = BC) ->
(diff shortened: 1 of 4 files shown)


awalterschulze:
The current problem with this pull request is that the final break is not preserved

Given
```
Foo andalso
    Bar andalso
    Baz
```
The break after `Bar` is not preserved and it is formatted to
```
Foo andalso
    Bar andalso Baz
```

awalterschulze:
If wondering about rewrite_assoc, I made a mistake before and retested my logic
```erlang

rewrite_assoc(A, {B, C} = _BC) ->
    rewrite_assoc(rewrite_assoc(A, B), C);
rewrite_assoc(A, B) ->
    {A, B}.

assoc(Config) when is_list(Config) ->
    {a, b} = rewrite_assoc(a, b),
    {{a, b}, c} = rewrite_assoc(a, {b, c}),
    {{{a, b}, c}, d} = rewrite_assoc(a, {b, {c, d}}),
    {{{{a, b}, c}, d}, e} = rewrite_assoc(a, {b, {c, {d, e}}}),
    {{{{{a, b}, c}, d}, e}, f} = rewrite_assoc(a, {b, {c, {d, {e, f}}}}).
```

awalterschulze:
The newline for 

```
Foo andalso
    Bar andalso
    Baz
```

is now preserved, given a new check `has_break_between_op` that knows that the op tree has been rewritten

awalterschulze:
The new problem is that the following

```
"One * (Two + Three + Four) * Five"
```

Is being formatted as

```
One *
    (Two + Three +
        Four) * Five
```

which in turn is being formatted as

```
One *
    (Two + Three +
        Four) *
    Five
```

which is being formatted as

```
One *
    (Two + Three +
        Four) * Five
```

awalterschulze:
Added test cases for #161

src/erlfmt_format.erl
```diff
@@ -307,14 +307,23 @@ binary_op_to_algebra('/', _Meta, {atom, _, _} = Left, {integer, _, _} = Right) -
...
     union_to_algebra(Meta, Left, Right);
-binary_op_to_algebra(Op, Meta0, Left, Right) ->
+binary_op_to_algebra(Op, Meta0, Left0, Right0) ->
     %% don't print parens and comments twice - expr_to_algebra took care of it already
     Meta = erlfmt_scan:delete_annos([parens, pre_comments, post_comments], Meta0),
+    {op, _Meta, Op, Left, Right} = rewrite_assoc(Op, Meta, Left0, Right0),
     binary_op_to_algebra(Op, Meta, Left, Right, ?INDENT).
 
+rewrite_assoc(Op, MetaA, A, {op, MetaBC, Op, B, C} = BC) ->
+    case erlfmt_scan:get_anno(parens, MetaBC, false) of
```
michalmuskala, line 317: I wonder if we need to check for comments in here. How do they interact with this flipping?

src/erlfmt_format.erl
```diff
@@ -307,14 +307,23 @@ binary_op_to_algebra('/', _Meta, {atom, _, _} = Left, {integer, _, _} = Right) -
...
+        true -> {op, MetaA, Op, A, BC};
+        _ -> rewrite_assoc(Op, MetaA, rewrite_assoc(Op, MetaBC, A, B), C)
```
michalmuskala, line 319: I believe we could remove the complexity in `has_break_between_op` if we tracked the locations here and amended `Meta` nodes. Something like:

```suggestion
        _ -> 
            AB = rewrite_assoc(Op, update_meta_location(MetaBC, A, B), A, B),
            rewrite_assoc(Op, update_meta_location(MetaA, AB, C), AB, C)
```
Given the following helper function
```erlang
update_meta_location(Meta, First, Last) ->
    Meta#{
        location := erlfmt_scan:get_anno(First, location),
        end_location := erlfmt_scan:get_anno(Last, end_location)
    }.
```

test/erlfmt_format_SUITE.erl
```diff
@@ -545,6 +547,200 @@ binary_operator(Config) when is_list(Config) ->
...
+    ),
+    ?assertSame(
+        "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA orelse\n"
+        "    BBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBBB andalso\n"
+        "        CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC.\n"
+    ),
+    ?assertSame(
+        "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa andalso\n"
+        "    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb orelse\n"
+        "    cccccccccccccccccccccccccccccccccc.\n"
```
michalmuskala, line 736: I understand why it's formatted like this, but it's not very clear when just taking a brief look. I wish we could add parens in cases like this, it would make it much, much clearer:
```erlang
(aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa andalso
    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb) orelse
    cccccccccccccccccccccccccccccccccc
```

awalterschulze: I can add the test with parens, but other than that I think it is important to also test without parens

src/erlfmt_format.erl
awalterschulze: I will try this, great idea :)

src/erlfmt_format.erl
awalterschulze: I'll add some more tests, good idea

test/erlfmt_format_SUITE.erl
```diff
@@ -545,6 +547,185 @@ binary_operator(Config) when is_list(Config) ->
...
+        "    equivalent(L4, R4)\n"
+    ),
+    ?assertFormat(
+        "the:ever(funny) andalso begin blocksof:my(code), that:appear(between), your:boolean(expressions) end andalso are:terrible(things, to, debug)\n",
+        "the:ever(funny) andalso\n"
+        "    begin\n"
+        "        blocksof:my(code),\n"
+        "        that:appear(between),\n"
+        "        your:boolean(expressions)\n"
+        "    end andalso are:terrible(things, to, debug)\n"
```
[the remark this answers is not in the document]
awalterschulze: What format would you prefer?

src/erlfmt_format.erl
awalterschulze: This worked out very well, thank you

test/erlfmt_format_SUITE.erl
awalterschulze: I had added the extra test, but don't know what else to do

src/erlfmt_format.erl
awalterschulze: Testing for comments was a great idea, it definitely looks better than without checking.
You can see the result in two tests I added at the bottom.
Not sure if it is ideal, but it could be acceptable.

test/erlfmt_format_SUITE.erl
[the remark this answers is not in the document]
michalmuskala: Probably with a break after `end`, but I'm not really sure. I checked and Elixir does the same as we do in this case

michalmuskala (approved):
Looks great!

Seems like there's some issue with the commits, though, since the `record` changes look like included in this PR too

awalterschulze:
I hope the squash merge takes care of it.
I tried to rebase on master and push it back to this branch, I think this is where the issue came in.

PR #269, opened 2021-03-31
better error reporting for not_equivalent and tests

awalterschulze:
Steal better error reporting from https://github.com/WhatsApp/erlfmt/compare/receive_clauses?expand=1


michalmuskala (approved):
😍

PR #271, opened 2021-04-01
Preserve comments in receive after after

awalterschulze:
Fixes https://github.com/WhatsApp/erlfmt/issues/125

Before this diff the following test would have passed where the comment is moved from before the `after` to above the `after`

```erlang
?assertFormat(
        "receive\n"
        "after\n"
        "    % before zero\n"
        "    0 -> ok\n"
        "end\n",
        "receive\n"
        "% before zero\n"
        "after 0 -> ok\n"
        "end\n"
    ),
```

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -796,8 +796,8 @@ receive_after_to_algebra(Expr, [HBody | _] = Body) ->
                     <<"after">>,
                     line(),
                     ExprD,
-                    <<" ->">>,
-                    nest(concat(break(<<"">>), BodyD), ?INDENT, always)
+                    <<" -> ">>,
+                    nest(BodyD, ?INDENT, break)
                 ])
         end,
     group(nest(Doc, ?INDENT)).
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -777,13 +777,22 @@ single_clause_spec_to_algebra(Name, {spec_clause, CMeta, Head, Body, Guards}) ->
     {args, AMeta, Args} = Head,
     clauses_to_algebra([{spec_clause, CMeta, {call, AMeta, Name, Args}, Body, Guards}]).
 
-receive_after_to_algebra(Expr, [HBody | _] = Body) ->
+receive_after_to_algebra(Expr, Body) ->
     ExprD = expr_to_algebra(Expr),
     BodyD = block_to_algebra(Body),
-    MaybeForce = maybe_force_breaks(has_break_between(Expr, HBody)),
-    HeadD = group(concat([<<"after">>, break(), ExprD, <<" ">>])),
-    Doc = nest(concat([<<"->">>, break(), MaybeForce, BodyD]), ?INDENT),
-    group(concat(HeadD, Doc)).
+    Nest = fun(List) -> group(nest(concat(List), ?INDENT, break)) end,
+    concat([
+        <<"after">>,
+        Nest([
+            break(),
+            ExprD,
+            <<" ->">>,
+            Nest([
+                break(),
+                BodyD
+            ])
+        ])
+    ]).
 
 try_to_algebra(Body, OfClauses, CatchClauses, After) ->
     Clauses =
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -780,27 +780,10 @@ single_clause_spec_to_algebra(Name, {spec_clause, CMeta, Head, Body, Guards}) ->
 receive_after_to_algebra(Expr, [HBody | _] = Body) ->
     ExprD = expr_to_algebra(Expr),
     BodyD = block_to_algebra(Body),
-    Doc =
-        case comments(Expr) of
-            {[], []} ->
-                concat([
-                    <<"after ">>,
-                    ExprD,
-                    <<" ->">>,
-                    break(),
-                    maybe_force_breaks(has_break_between(Expr, HBody)),
-                    BodyD
-                ]);
-            _ ->
-                concat([
-                    <<"after">>,
-                    line(),
-                    ExprD,
-                    <<" -> ">>,
-                    BodyD
-                ])
-        end,
-    group(nest(Doc, ?INDENT)).
+    MaybeForce = maybe_force_breaks(has_break_between(Expr, HBody)),
+    HeadD = group(concat([<<"after">>, break(), ExprD, <<" ">>])),
+    Doc = nest(concat([<<"->">>, break(), MaybeForce, BodyD]), ?INDENT),
+    group(concat(HeadD, Doc)).
 
 try_to_algebra(Body, OfClauses, CatchClauses, After) ->
     Clauses =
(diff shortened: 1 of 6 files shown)


test/erlfmt_format_SUITE.erl
```diff
@@ -2574,13 +2574,31 @@ receive_expression(Config) when is_list(Config) ->
...
+        "receive\n"
+        "after\n"
+        "    % before zero\n"
+        "    0 -> ok\n"
+        "end\n",
+        "receive\n"
+        "after\n"
+        "    % before zero\n"
+        "    0 ->\n"
+        "        ok\n"
```
michalmuskala, line 2599: Do we force this to break if there's a comment? I don't think we should

test/erlfmt_SUITE_data/simple_comments.erl
```diff
@@ -160,7 +160,7 @@ call() ->
     end,
     receive
         ok -> ok
-    %% comment 5
+        %% comment 5
```
michalmuskala, line 163: Hmm... This is a bit unfortunate, though I understand why it's happening. I think the case of adding a comment to after like it was here before would be quite common

awalterschulze: Agreed

test/erlfmt_format_SUITE.erl
```diff
@@ -2591,12 +2608,53 @@ receive_expression(Config) when is_list(Config) ->
...
+        "receive\n"
+        "    1 -> ok\n"
+        "    %% after receive\n"
+        "after\n"
+        "    %% before zero\n"
+        "    0 -> a(\n"
+        "            b,\n"
+        "            c\n"
+        "        )\n"
+        "        %% after after for receive\n"
```
michalmuskala: Oh, I missed this one before.

I don't think we should do that. We don't do it for regular clauses. Is there a particular reason why here it's desirable?

awalterschulze: This is a new test

awalterschulze: You didn't miss it before.

test/erlfmt_format_SUITE.erl
awalterschulze: No longer forced

src/erlfmt_format.erl
```diff
@@ -776,11 +778,29 @@ single_clause_spec_to_algebra(Name, {spec_clause, CMeta, Head, Body, Guards}) ->
...
+                    maybe_force_breaks(has_break_between(Expr, HBody)),
+                    BodyD
+                ]);
+            _ ->
+                concat([
+                    <<"after">>,
+                    line(),
+                    ExprD,
+                    <<" -> ">>,
+                    BodyD
```
awalterschulze, line 800: Adding the break here, always breaks
```
break(),
maybe_force_breaks(has_break_between(Expr, HBody)),
```

test/erlfmt_format_SUITE.erl
michalmuskala: I don't think this is correct. i just tried it, and for example with a concat node as the value, it does something like this:

```erl
receive
after
    %% foo
    0 -> "abc"
    "def"
end.
```

src/erlfmt_format.erl
awalterschulze: Adding just a `break()` here also, always breaks

test/erlfmt_format_SUITE.erl
awalterschulze: Good catch. I have added it as a test.

test/erlfmt_format_SUITE.erl
```diff
@@ -2574,13 +2574,47 @@ receive_expression(Config) when is_list(Config) ->
...
+        "receive\n"
+        "after\n"
+        "    %% foo\n"
+        "    0 ->\n"
+        "        \"abc\"\n"
+        "        \"def\"\n"
+        "end.\n",
+        "receive\n"
+        "after\n"
+        "%% foo\n"
```
awalterschulze, line 2612: This is a downside of the current solution.
Where the clause is not indented.

The problem with indenting the clause is then when we loose the `0 ->` on the same line as the after.
```
?assertFormat(
        "receive\n"
        "    1 -> ok\n"
        "after 0 ->\n"
        "    ok\n"
        "end\n",
        "receive\n"
        "    1 -> ok\n"
        "after\n"
        "    0 ->\n"
        "        ok\n"
        "end\n"
    ),
```

awalterschulze: I am wondering if the current solution is an okay compromise, at least the comment stays in place.

awalterschulze: For reference this would be the code for the formatting in the top case
```
receive_after_to_algebra(Expr, [HBody | _] = Body) ->
    ExprD = expr_to_algebra(Expr),
    BodyD = block_to_algebra(Body),
    MaybeForce = maybe_force_breaks(has_break_between(Expr, HBody)),
    ArrowBody = group(nest(concat([<<"->">>, break(), MaybeForce, BodyD]), ?INDENT)),
    Doc = group(nest(concat([break(), ExprD, <<" ">>, ArrowBody]), ?INDENT, break)),
    concat(<<"after">>, Doc).
```

awalterschulze: The groups break upward, so all the nesting and breaks are triggered

test/erlfmt_format_SUITE.erl
awalterschulze: These are both fixed now, but see new issues below.

src/erlfmt_recomment.erl
```diff
@@ -217,13 +217,19 @@ insert_nested({'case', Meta, Expr0, Clauses0}, Comments0) ->
...
 insert_nested({'receive', Meta, Clauses0}, Comments0) ->
-    Clauses = insert_expr_container(Clauses0, Comments0),
+    {Clauses, []} = insert_expr(Clauses0, Comments0),
     {{'receive', Meta, Clauses}, []};
-insert_nested({'receive', Meta, Clauses0, AfterExpr0, AfterBody0}, Comments0) ->
-    {Clauses, Comments1} = insert_expr_list(Clauses0, Comments0),
-    {AfterExpr, Comments2} = insert_expr(AfterExpr0, Comments1),
-    AfterBody = insert_expr_container(AfterBody0, Comments2),
-    {{'receive', Meta, Clauses, AfterExpr, AfterBody}, []};
+insert_nested({'receive', Meta, empty, {after_clause, _, _, _} = After0}, Comments0) ->
```
michalmuskala, line 222:
```suggestion
insert_nested({'receive', Meta, empty, After0}, Comments0) ->
```
This can only be an `after_clause` node. The match is confusing compared to the next clause

test/erlfmt_format_SUITE.erl
```diff
@@ -2490,11 +2493,15 @@ receive_expression(Config) when is_list(Config) ->
...
-    ?assertSame(
+    ?assertFormat(
         "receive\n"
         "    1 -> ok\n"
         "after 0 ->\n"
         "    ok\n"
+        "end\n",
+        "receive\n"
+        "    1 -> ok\n"
+        "after 0 -> ok\n"
```
awalterschulze, line 2504: Here we loose the newline preserving with this algorithm.

michalmuskala (approved):

Issue #272, opened 2021-04-06
Preserve newlines in specs


michalmuskala:
Preserve the newline after `->` in specs.

Today:
```erl
-spec foo() ->
    atom().
```

would be re-formatted to:
```erl
-spec foo() -> atom().
```

After this change it would stay as in the original.

This change would be consistent with how we preserve newlines in function clauses, case clauses and other clause-like constructs.

PR #273, opened 2021-04-12
preserve newline in spec without guards

awalterschulze:
Possibly fixes https://github.com/WhatsApp/erlfmt/issues/272

This makes the following a mirror of each other:
`clause_to_algebra({clause, _Meta, Head, empty, Body}) ->`
`clause_to_algebra({spec_clause, _Meta, Head, Body, empty}) ->`

This means we remove a `group` around `nest`, which when the spec parameters breaks the space after `->` return also breaks.

An alternative would be to keep the group, but add 
`MaybeForce = maybe_force_break(has_break_between(Head, Body))`
`group(nest(concat([MaybeForce, <<"->">>, break(<<" ">>), BodyD]), ?INDENT))`
But then maybe we should do the same for the other two `clause_to_algebra`s for consistency?

### DIFF
--- a/test/erlfmt_SUITE_data/comments.erl.formatted
+++ b/test/erlfmt_SUITE_data/comments.erl.formatted
@@ -31,7 +31,9 @@ has_fanciness([]) ->
     false.
 
 %% comment
-bar(X) when is_list(X) ->
+bar(X) when
+    is_list(X)
+->
     ok;
 %% comment
 bar(X) when is_atom(X) ->
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -722,12 +722,14 @@ clause_to_algebra({clause, Meta, Head, Guards, Body}) ->
     GuardsD = expr_to_algebra(Guards),
     BodyD = block_to_algebra(Meta, Body),
     MaybeForceHeadGuards = maybe_force_breaks(has_break_between(Head, Guards)),
+    MaybeForceGuardsBody = maybe_force_breaks(has_break_between(Guards, Body)),
     Nested = fun(Doc) -> nest(concat(break(<<" ">>), Doc), ?INDENT) end,
     concat(
         space(HeadD, <<"when">>),
         group(
             concat(
                 group(concat([MaybeForceHeadGuards, Nested(GuardsD), break(<<" ">>), <<"->">>])),
+                MaybeForceGuardsBody,
                 Nested(BodyD)
             )
         )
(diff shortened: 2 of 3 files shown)


awalterschulze:
Found another bug that

```
% ?assertFormat(
    %     "-spec foo(Int) ->\n"
    %     "    atom() when\n"
    %     "    Int :: integer().\n",
    %     "-spec foo(Int) -> atom() when\n"
    %     "    Int :: integer().\n"
    % ),
```

This format does not stabilize.

We cycle between
`-spec foo(Int) -> atom() when Int :: integer().`
and the result in `assertFormat`

awalterschulze:
> Found another bug
> 
> ```
> % ?assertFormat(
>     %     "-spec foo(Int) ->\n"
>     %     "    atom() when\n"
>     %     "    Int :: integer().\n",
>     %     "-spec foo(Int) -> atom() when\n"
>     %     "    Int :: integer().\n"
>     % ),
> ```
> 
> This format does not stabilize.
> 
> We cycle between
> `-spec foo(Int) -> atom() when Int :: integer().`
> and the result in `assertFormat`

This is fixed in https://github.com/WhatsApp/erlfmt/pull/273/commits/ff36b47fe154422d25a0c67221d5f88ba3d22f31

awalterschulze:
Currently the newline is only preserved in specs without guards, see

```
?assertFormat(
        "-spec foo(Int) ->\n"
        "    atom() when\n"
        "    Int :: integer().\n",
        "-spec foo(Int) -> atom() when Int :: integer().\n"
    ),
```

But this is consistent with how other clauses are handled.
@michalmuskala do we want to preserve newlines in these cases too?

test/erlfmt_SUITE_data/comments.erl.formatted
```diff
@@ -31,7 +31,9 @@ has_fanciness([]) ->
     false.
 
 %% comment
-bar(X) when is_list(X) ->
+bar(X) when
```
awalterschulze, line 34: The original was
```
bar(X) when %% comment
            is_list(X) ->
    ok;
```
which has a newline that is now preserved

awalterschulze:
> Currently the newline is only preserved in specs without guards, see
> 
> ```
> ?assertFormat(
>         "-spec foo(Int) ->\n"
>         "    atom() when\n"
>         "    Int :: integer().\n",
>         "-spec foo(Int) -> atom() when Int :: integer().\n"
>     ),
> ```
> 
> But this is consistent with how other clauses are handled.
> @michalmuskala do we want to preserve newlines in these cases too?

I have added `MaybeForce`'s for clauses with guards

michalmuskala (approved):

PR #274, opened 2021-04-12
Fix end_location in raw_string anno

gomoripeti:
I noticed that for unparseable forms erlfmt parser returns incorrect end_location. The end_location is not used by erlfmt itself, but it can be useful for other tools. I'm unsure where a new testcase would fit as this internal thing does not affect formatting at all.

An example: (end column is the number of tokens (8) instead of the correct value 26)
```
1> {ok, Forms, _Warnings} = erlfmt:read_nodes_string("nofile", "f() -> \"very long string\""), Forms.
[{raw_string,#{end_location => {1,8},location => {1,1}},
             "f() -> \"very long string\""}]
2> length("f() -> \"very long string\"").
25
```

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -317,8 +317,8 @@ contains_ignore_comment({comment, _Loc, Comments}) ->
     ).
 
 node_string(Cont) ->
-    {String, Anno} = erlfmt_scan:last_node_string(Cont),
-    {raw_string, Anno, string:trim(String, both, "\n")}.
+    {String, Anno} = erlfmt_scan:last_node_string_trimmed(Cont),
+    {raw_string, Anno, String}.
 
 -spec format_nodes([erlfmt_parse:abstract_form()], pos_integer()) -> [unicode:chardata()].
 format_nodes([], _PrintWidth) ->
@@ -467,8 +467,8 @@ nodes_in_range(Nodes, StartLocation, EndLocation) ->
 
 node_intersects_range(Node, StartLocation, EndLocation) ->
     {Start, End} = get_location_range(Node),
-    ((Start < StartLocation) and (End >= StartLocation)) or
-        ((Start >= StartLocation) and (Start =< EndLocation)).
+    ((Start < StartLocation) and (End > StartLocation)) or
+        ((Start >= StartLocation) and (Start < EndLocation)).
 
 get_possible_locations([Option1, Option2 | _], Location, GetLoc) ->
     case GetLoc(Option1) of
--- a/src/erlfmt_scan.erl
+++ b/src/erlfmt_scan.erl
@@ -15,7 +15,14 @@
 
 -include("erlfmt_scan.hrl").
 
--export([io_node/1, string_node/1, continue/1, last_node_string/1]).
+-export([
+    io_node/1,
+    string_node/1,
+    continue/1,
+    last_node_string/1,
+    last_node_string_trimmed/1
+]).
+
 -export([
     put_anno/3,
     merge_anno/2,
@@ -180,6 +187,13 @@ last_node_string(#state{original = [First | _] = Tokens}) ->
     Location = erl_scan:location(First),
     {String, token_anno([{text, String}, {location, Location}])}.
 
+-spec last_node_string_trimmed(state()) -> {unicode:chardata(), anno()}.
+last_node_string_trimmed(#state{original = [First | _] = Tokens}) ->
+    String0 = stringify_tokens(Tokens),
+    String = string:trim(String0, both, "\n"),
+    Location = erl_scan:location(First),
+    {String, token_anno([{text, String}, {location, Location}])}.
+
 has_new_line({white_space, _, [$\n | _]}) -> true;
 has_new_line(_) -> false.
 
@@ -211,8 +225,9 @@ drop_initial_white_space([{white_space, _, _} | Rest]) ->
 drop_initial_white_space(Rest) ->
     Rest.
 
+-spec stringify_tokens([erl_scan:token()]) -> string().
 stringify_tokens(Tokens) ->
-    lists:map(fun erl_scan:text/1, Tokens).
+    lists:flatmap(fun erl_scan:text/1, Tokens).
 
 -spec split_tokens([erl_scan:token()], [erl_scan:token()]) ->
     {[token()], [erl_scan:token()], [comment()], [token()]}.
--- a/test/erlfmt_SUITE.erl
+++ b/test/erlfmt_SUITE.erl
@@ -81,7 +81,8 @@
     contains_pragma/1,
     insert_pragma/1,
     overlong_warning/1,
-    do_not_crash_on_bad_record/1
+    do_not_crash_on_bad_record/1,
+    raw_string_anno/1
 ]).
 
 suite() ->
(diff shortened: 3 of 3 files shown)


awalterschulze:
Thanks for helping to fix this, really appreciate it :)

I think a test in `erlfmt_SUITE.erl` in the `parser` group would be helpful.

Also a lot of tests are breaking, probably because of a missing flatten somewhere else before comparing with assert.

gomoripeti:
At this point I would consider this PR as just an issue ticket, as I am not familiar with the code enough to know the best way to address the issue.
What I figured out is that in `erlfmt:node_string/1` the string returned by `erlfmt_scan:last_node_string(Cont)` is trimmed, but the Anno will contain start/end location from the original string.
https://github.com/WhatsApp/erlfmt/blob/master/src/erlfmt.erl#L319-L321
When formatting `has_empty_line_between/2` might think there is no empty line based on the annotation, but because of the trimming there should be an empty line inserted.
https://github.com/WhatsApp/erlfmt/blob/master/src/erlfmt.erl#L343
Though I can hardly think of a scenario where this would cause a problem in formatting.
A contrived example for inconsistency:
```
1> erlfmt:format_string("foo?M -> ok.\ng?M ->ok.\n", []).
{ok,"foo?M -> ok.\ng?M ->ok.\n", [...]}
2> erlfmt:format_string("foo?M ->\n ok.\ng?M ->ok.\n", []).
{ok,"foo?M ->\n ok.\n\ng?M ->ok.\n", [...]}
```
In the second case there is a new line inserted between the two unparsable forms but not in the first case

Please feel free to close this PR and apply any change (if any) that you feel appropriate

awalterschulze:
When forms are not valid, we don't want to crash and reproduce the input.
After that I think we find it acceptable to insert a single empty line in some cases.

Does this breaks some use case?

gomoripeti:
I gave it another stab, and it's now ready for review.

I can't reproduce the checkfmt failure locally, so I dont know what the problem could be.

awalterschulze:
Try running `make fmt`, it should reformat the code using `erlfmt` and fix the `checkfmt` failure.

src/erlfmt.erl
```diff
@@ -498,9 +498,9 @@ check_line_lengths(FileName, Width, String, {FirstLineNo, _}) ->
     LastLineNo = FirstLineNo + length(Lines) - 1,
     [
         {FileName, LineNo, ?MODULE, {long_line, string:length(Line), Width}}
-     || {LineNo, Line} <- lists:zip(
-            lists:seq(FirstLineNo, LastLineNo),
-            Lines
-        ),
-        string:length(Line) > Width
+        || {LineNo, Line} <- lists:zip(
```
awalterschulze, line 501: This is an old format

src/erlfmt_algebra.erl
```diff
@@ -443,10 +443,11 @@ format(Doc, Width) when ?is_doc(Doc) andalso (Width == infinity orelse Width >=
 -spec fits(Width :: integer(), Column :: integer(), HasBreaks :: boolean(), Entries) ->
     boolean()
 when
-    Entries :: maybe_improper_list(
-        {integer(), mode(), doc()},
-        {tail, boolean(), Entries} | []
-    ).
+    Entries ::
```
awalterschulze, line 446: This is also an old format

awalterschulze (commented):
I don't know if you are reformatting using the newest version of erlfmt

src/erlfmt.erl
gomoripeti: Thank you for the heads up, this must be the issue, somehow locally I'm running with an old formatter, will try to figure it out.

gomoripeti: turns out I had an `ebin` dir in the root of my repo with old compiled beam files. After removing all is well. Sorry about the noise.

awalterschulze: This happens to me so much, no worries.
I run `rm -rf _build` probably much more than is required.

src/erlfmt_scan.erl
```diff
@@ -180,6 +187,13 @@ last_node_string(#state{original = [First | _] = Tokens}) ->
     Location = erl_scan:location(First),
     {String, token_anno([{text, String}, {location, Location}])}.
 
+-spec last_node_string_trimmed(state()) -> {unicode:chardata(), anno()}.
```
awalterschulze, line 190: I wonder what happens when `last_node_string_trimmed` is used instead of `last_node_string` for all cases ?

src/erlfmt.erl
```diff
@@ -467,8 +467,8 @@ nodes_in_range(Nodes, StartLocation, EndLocation) ->
 
 node_intersects_range(Node, StartLocation, EndLocation) ->
     {Start, End} = get_location_range(Node),
-    ((Start < StartLocation) and (End >= StartLocation)) or
-        ((Start >= StartLocation) and (Start =< EndLocation)).
+    ((Start < StartLocation) and (End > StartLocation)) or
```
awalterschulze, line 470: Just interested to know, why was this change required?

awalterschulze (commented):
This seems like a very useful change.
Thanks for making the effort.
Just some minor questions.

src/erlfmt_scan.erl
gomoripeti: that's a good question. I don't fully understand the main read_nodes loop, so I'm not sure if trailing new lines can happen there and what affect of trimming them would have on formatting.

awalterschulze: In the interest of keeping the API small, what happens if you simply replace it everywhere, does anything break when you run `make test`?

src/erlfmt.erl
gomoripeti: This was an intermediate fix for my previous attempt, to fix one of the format_range test cases. Without trimming the unparseable ranges could be side by side (ie the end location of first one can be the same as the start location of the following). Now that the raw_strings are trimmed and the end_locations are matching this cannot happen, but I thought it still makes sense to make this function theoretically correct.

src/erlfmt_scan.erl
gomoripeti: there are two cases failing, these are the diffs

```
$ diff output/no_pragma.erl test/erlfmt_SUITE_data/no_pragma.erl
1c1
< -module(no_pragma). % "שלום עולם" non-unicode trailing comment
\ No newline at end of file
---
> -module(no_pragma). % "שלום עולם" non-unicode trailing comment
```
```
$ diff output/unicode.erl test/erlfmt_SUITE_data/unicode.erl
1c1,3
< -module(unicode).%% we read ahead by one form, so we need this to test properly read_rest with what follows
---
> -module(unicode).
>
> %% we read ahead by one form, so we need this to test properly read_rest with what follows
```

awalterschulze: Interesting, these are both at the top, when we are still finding the pragma, so this might be fixable, but I think we can tackle that in a following diff.
I think this is ready to merge, but I'll wait for @michalmuskala to weigh in.

awalterschulze (approved):

michalmuskala (approved):

awalterschulze:
@michalmuskala and I talked about this.
We will work on removing `last_node_string_trimmed` in favour of `last_node_string` after fixing our usage of `last_node_string`, but do not think you should be blocked on us having to fix our code.  Everything you did in this pull request looks great.  Thanks again for your help.

Issue #277, opened 2021-04-20
Add new pragma alternative


awalterschulze:
`%% @format` seems to conflict with edoc
We need to add an alternative pragma and support both for backwards compatibility
We also need an option to update all `%% @format` pragmas to the new pragma
I guess this can just be using `--insert-pragma` on a file that already has an old pragma

awalterschulze:
See conversation in https://github.com/AdRoll/rebar3_format/issues/243

PR #281, opened 2021-04-28
comment the @format pragma to avoid edoc warnings

awalterschulze:
Fixes https://github.com/WhatsApp/erlfmt/issues/277

Some context: Marc-Andre has shown me an analyses of triple comments that removes the bias in my analyses, that shows that they are actually popular in erlang before the module attribute.  We will work on pushing this analyses to the repo, but I think we should adopt this style in insert pragma

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -200,7 +200,8 @@ replace_pragma_comment_blocks(_Prefix, []) ->
     [];
 replace_pragma_comment_blocks(Prefix, [{comment, Loc, Comments} | Rest]) ->
     case replace_pragma_comment_block(Prefix, Comments) of
-        [] -> replace_pragma_comment_block(Prefix, Rest);
+        [] ->
+            replace_pragma_comment_block(Prefix, Rest);
         CleanComments ->
             {Prefix0, _} = string:take(lists:last(CleanComments), "%"),
             [{comment, Loc, CleanComments} | replace_pragma_comment_block(Prefix0, Rest)]
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -147,8 +147,9 @@ insert_pragma_node(Node) ->
                 [{comment, #{location => Loc, end_location => Loc}, ["%%% % @format", ""]}];
             _ ->
                 {comment, Loc, LastComments} = lists:last(PreComments),
+                {Prefix, _} = string:take(lists:last(LastComments), "%"),
                 lists:droplast(PreComments) ++
-                    [{comment, Loc, LastComments ++ ["%%% % @format"]}]
+                    [{comment, Loc, LastComments ++ [Prefix ++ " % @format"]}]
         end,
     erlfmt_scan:put_anno(pre_comments, NewPreComments, Node).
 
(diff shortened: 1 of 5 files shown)


src/erlfmt.erl
```diff
@@ -144,11 +144,11 @@ insert_pragma_node(Node) ->
         case PreComments of
             [] ->
                 Loc = erlfmt_scan:get_anno(location, Node),
-                [{comment, #{location => Loc, end_location => Loc}, ["%% @format", ""]}];
+                [{comment, #{location => Loc, end_location => Loc}, ["%%% % @format", ""]}];
             _ ->
                 {comment, Loc, LastComments} = lists:last(PreComments),
                 lists:droplast(PreComments) ++
-                    [{comment, Loc, LastComments ++ ["%% @format"]}]
+                    [{comment, Loc, LastComments ++ ["%%% % @format"]}]
```
michalmuskala, line 151: I feel like the proper way to do this would be to detect whether to use `%%` or `%%%`, In particular something like:

```erl
{Prefix, _} = string:take(lists:last(LastComment), "%"),
[{comment, Loc, LastComments ++ [Prefix ++ " % @format"]}]
```

I don't have a preference for what should be the default in case there weren't any comments already.

src/erlfmt.erl
```diff
@@ -188,6 +188,28 @@ remove_pragma_comment_block([Head | Tail]) ->
...
+        [] -> replace_pragma_comment_block(Rest);
+        CleanComments -> [{comment, Loc, CleanComments} | replace_pragma_comment_block(Rest)]
+    end.
+
+replace_pragma_comment_block([]) ->
+    [];
+replace_pragma_comment_block(["%% @format" | Tail]) ->
+    ["%%% % @format" | Tail];
+replace_pragma_comment_block([Head | Tail]) ->
+    [Head | replace_pragma_comment_block(Tail)].
```
michalmuskala, line 211: I'm not convinced about forcing the triple comment. What if the surrounding comment uses only double? It would be strange to end up with something like:

```erl
%% Some licence info
%%
%%% % @format
%%
%% @doc ....
```

awalterschulze: Yes we could replace it with the number of percentages that was there.

src/erlfmt.erl
awalterschulze: We could, but we still need a default for the case where there are no comments yet and in this case I think we should do 3 comments, given this new analyses: https://github.com/WhatsApp/erlfmt/pull/283

awalterschulze: I have added your suggestion :)

src/erlfmt.erl
awalterschulze: I have also address this with a default of 3, but if some other number is detected we prefer that.

src/erlfmt.erl
```diff
@@ -188,6 +189,32 @@ remove_pragma_comment_block([Head | Tail]) ->
...
+    PostComments = replace_pragma_comment_blocks("%%%", PostComments0),
+    Node = erlfmt_scan:put_anno(pre_comments, PreComments, Node0),
+    erlfmt_scan:put_anno(post_comments, PostComments, Node).
+
+replace_pragma_comment_blocks(_Prefix, []) ->
+    [];
+replace_pragma_comment_blocks(Prefix, [{comment, Loc, Comments} | Rest]) ->
+    case replace_pragma_comment_block(Prefix, Comments) of
+        [] ->
+            replace_pragma_comment_block(Prefix, Rest);
```
michalmuskala, line 204: Can this happen? I don't think we ever produce empty comment blocks and `replace_pragma_comment_block/2` never drops any list elements

michalmuskala (approved):

src/erlfmt.erl
awalterschulze: Good catch, I updated the code.

PR #290, opened 2021-05-17
New assert macros for smarter lists or maps mismatch diagnostic.

slaykachu:
When 2 lists aren't equal, instead of printing both,
?assertListsEqual will only show differences.
Similarily for ?assertMapsEqual. Caveat:
this latter one stops at first mismatch.

This is quite useful when structures are huge.

erlfmt_SUITE has been refactored to demonstrate use.


test/erlfmt_SUITE.erl
```diff
@@ -1154,25 +1155,34 @@ snapshot_insert_pragma_with(Config) when is_list(Config) ->
...
+    case erlfmt:format_string(Original, Options) of
+        {ok, Formatted, _} -> ok;
+        {skip, _} -> ok;
+        {ok, Other, _} ->
+           % Split by lines.
+           Formatted2 = string:lexemes(Formatted, [$\n]),
+           Other2 = string:lexemes(Other, [$\n]),
+           % We already know they are not equal,
+           % this macro gives a better diagnostic.
+           ?assertListEqual(Formatted2, Other2);
```
awalterschulze, line 1181: Ah okay I really like this idea, it has been really tough to read snapshot test differences.  Thanks for doing work to improve the current debugging process.  This has been a bit of guess work in the past.

src/assert_diagnostic.erl
```diff
@@ -0,0 +1,46 @@
...
+%% See assert_diagnostic.hrl.
+
+-module(assert_diagnostic).
+
+-export([
+   check_length/2,
+   check_elements/3
+]).
+
+-define(FORMAT(T, Args), lists:flatten(io_lib:format(T, Args))).
```
awalterschulze, line 24: We have tried to write as little macros as possible.  Is it possible to remove this one, I understand we might need the `assertListEqual` macro, but lets remove the rest that we do not have to have and rather use functions.

src/assert_diagnostic.hrl
```diff
@@ -0,0 +1,47 @@
...
+        B -> ok;
+        _ -> Errs = assert_diagnostic:check_length(A, B) ++
+                    assert_diagnostic:check_elements(A, B, 1),
+             ?assert(false, Errs)
+   end
+).
+
+%% Naive report:
+%% * Stop after first mismatch rather than collecting all differences.
+-define(assertMapsEqual(A, B),
```
awalterschulze, line 40: I don't think this is ever used.

test/assert_diagnostic_SUITE.erl
```diff
@@ -0,0 +1,62 @@
...
+        ]}
+    ].
+
+all() ->
+    [{group, my_test_group}].
+
+
+%% Helper ensuring comparison of mismatching lists give the expected message.
+%% Implemented as a macro to have proper line report if the test case fails.
+-define(checkListEqualMessage(L0, L1, Msg),
```
awalterschulze, line 26: Could this be a function?

test/erlfmt_SUITE.erl
```diff
@@ -1154,25 +1155,34 @@ snapshot_insert_pragma_with(Config) when is_list(Config) ->
...
+snapshot_match(FormattedModule, Module, Config, Options) ->
     DataDir = ?config(data_dir, Config),
-    {ok, FormattedBin} = file:read_file(filename:join([DataDir, Module ++ ".formatted"])),
+    {ok, FormattedBin} = file:read_file(filename:join([DataDir, FormattedModule])),
     Formatted = unicode:characters_to_list(FormattedBin),
     {ok, OriginalBin} = file:read_file(filename:join([DataDir, Module])),
     Original = unicode:characters_to_list(OriginalBin),
-    ?assertMatch({ok, Formatted, _}, erlfmt:format_string(Original, [])).
+    case erlfmt:format_string(Original, Options) of
+        {ok, Formatted, _} -> ok;
```
awalterschulze, line 1173: Maybe this should be `{ok, Original, _}` ?

test/assert_diagnostic_SUITE.erl
slaykachu: I'd rather not!
- that's a very local helper, not meant to be used anywhere else anyway.
- with a function, the failure message wouldn't point to the test itself (would give live in function body rather line in test body).

test/erlfmt_SUITE.erl
slaykachu: Why? You want to check the result of format_string against the reference file.
I think 'Expected' would be less ambiguous, but I kept the existing terminology to have the minimum diff possible!

src/assert_diagnostic.hrl
slaykachu: That's a mini library! Isn't it nice to know it's there and available?

michalmuskala:
This looks very useful. No more staring at almost the same strings 😅

However, I'm not sure I'm a big fan of placing this directly in `src` if we're only using it for testing. There's always the risk of some integrator using it and treating as accidental public API (since you can't "hide" modules in erlang). Furthermore, there are potential for module name clashes leading to issues for users - normally modules within an application are prefixed with the application name (something like `erlfmt_assert_diagnostics` in this case).

test/assert_diagnostic_SUITE.erl
awalterschulze: Okay fair "the failure message wouldn't point to the test itself" since we are only using it once, how about we inline it?
It would be really great if we can avoid macros as much as we can.

test/erlfmt_SUITE.erl
awalterschulze: I was looking at original code and not paying much attention.
I probably missed something.

Screen Shot 2021-05-17 at 5 08 15 PM

slaykachu: The code you are quoting is for `snapshot_same` where both the reference and the file to format are the same.
This is now expressed by passing `Module` twice to the helper function.

test/assert_diagnostic_SUITE.erl
slaykachu: Ok, turning that into a function!

slaykachu: Note: my point about failure location is mostly irrelevant here, since the unit tests are very short and only test one thing each (as it should be!).

src/assert_diagnostic.erl
slaykachu: Done!

slaykachu:
> However, I'm not sure I'm a big fan of placing this directly in src

Moved in test/ !

test/erlfmt_SUITE.erl
awalterschulze: Okay great, that makes sense.  My bad.

test/assert_diagnostic.erl
```diff
@@ -0,0 +1,66 @@
...
+check_elements([], _, _) ->
+    [];
+check_elements(_, [], _) ->
+    [];
+check_elements([H | T1], [H | T2], I) ->
+    check_elements(T1, T2, I + 1);
+check_elements([H1 | T1], [H2 | T2], I) ->
+    [
+        format(
+            "Item ~p differs:~n"
```
awalterschulze, line 60: Wouldn't this maybe be "Line ~p differs ..." ?

michalmuskala (approved):
Looks good to me!

awalterschulze (approved):
I would probably remove the unused `assertMapsEqual`, but otherwise this PR makes a BIG improvement to our debugging experience on large files which has been a problem for a long time.  Thank you for doing this.

awalterschulze:
@slaykachu let me know if you are happy and then I will `Squash and merge` it.

test/assert_diagnostic.erl
slaykachu: Depends on the context :)
Anyway right now it's not quite the line, since empty lines are removed by `string:lexemes`. 
This "bug" will be addressed in a subsequent commit.

PR #297, opened 2021-05-26
[Forrmat range] Re-inject formatted range in original source.

slaykachu:


test/erlfmt_SUITE.erl
```diff
@@ -1191,9 +1200,9 @@ assert_snapshot_match(Expected, Output) ->
         {skip, _} ->
             ok;
         {ok, Other, _} ->
-            % Split by lines.
-            Expected2 = string:lexemes(Expected, [$\n]),
-            Other2 = string:lexemes(Other, [$\n]),
+            % Split by lines (preserving empty lines).
+            Expected2 = string:split(Expected, "\n", all),
```
awalterschulze, line 1204: Not so familiar with the difference between `lexemes` and `split` how did you decide that this change was important?

awalterschulze: Oh I see

Screen Shot 2021-05-27 at 12 01 05 PM

Was this the reason?

src/erlfmt.erl
```diff
@@ -13,14 +13,18 @@
 %% limitations under the License.
 -module(erlfmt).
 
+-include_lib("stdlib/include/assert.hrl").
```
awalterschulze, line 16: You already mentioned this in the chat to me, but hopefully having a comment here makes it easy to tick off.

src/erlfmt.erl
```diff
@@ -239,6 +244,68 @@ format_file_range(FileName, StartLocation, EndLocation, Options) ->
...
+            New = replace_slice(AsList, Start, Len, FormattedAsList),
+            {ok, lists:join("\n", New), Info};
+        X ->
+            % Error: Leave as is.
+            X
+    end.
+
+% Like format_file_range, but only return the modified part.
+% Entry point for IDE integration.
+-spec format_file_range_extract(
```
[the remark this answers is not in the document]
slaykachu: Good point. It's just a residue of a previous (private) refactoring. Going to fix that.

src/erlfmt.erl
```diff
@@ -247,18 +314,18 @@ format_string_range(String, StartLocation, EndLocation, Options) ->
 format_enclosing_range(FileName, StartLocation, EndLocation, Options, Nodes, Warnings) ->
     case format_range(FileName, StartLocation, EndLocation, Options, Nodes, Warnings) of
         {options, PossibleRanges} ->
-            % Pick the largest range, so all intersected forms are covered.
+            % Pick the largest range, so all intersected forms are coverd.
```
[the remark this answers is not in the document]
slaykachu: Thanks! I see dead git ghosts.

test/erlfmt_SUITE.erl
slaykachu: The main reason is to preserve empty lines:
```
% Line 1
% Line 2

% Line 4
```
With lexemes, `% Line 4` would become the 3rd item of the list. So, in case of mismatch error, the index wouldn't correspond to the line number, which is confusing.

src/erlfmt.erl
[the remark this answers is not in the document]
slaykachu: Oh, actually it's still needed until I'm able to fix the dialyzer error!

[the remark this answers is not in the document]
slaykachu: I've added some comment about that. I'm postponing this cleanup, since current code might change anyway.

src/erlfmt.erl
slaykachu: Done!

src/erlfmt.erl
```diff
@@ -227,8 +231,9 @@ replace_pragma_comment_block(Prefix, [Head | Tail]) ->
     {ok, string(), [error_info()]}
     | {error, error_info()}.
 format_file_range(FileName, StartLocation, EndLocation, Options) ->
-    {ok, Nodes, Warnings} = file_read_nodes(FileName, ignore),
```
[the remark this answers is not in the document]
slaykachu: Trying to pass nodes complicate the code.
We lose symmetry, but we gain a nice flow: formatting the file is just "reading the file + formatting the string".

src/erlfmt.erl
```diff
@@ -19,8 +19,16 @@
     init/1,
     format_file/2,
     format_file_range/4,
+    format_file_range_extract/4,
     format_string/2,
     format_string_range/4,
+    format_string_range_extract/4,
+    % FIXME: We are exporting this function to quiet dialyzer.
```
awalterschulze, line 26: What exactly was the dialyzer error, sorry I didn't have time to check before
We should try to avoid big comments like this and try to fix it.

slaykachu: Cf https://github.com/WhatsApp/erlfmt/runs/2678261048
You can also reproduce locally by removing the export!

slaykachu: Apparently this function is ok. But when the code path is from `format_file_range` or `format_string_range`, dialyzer infers that passed `Result` cannot be `{ok, Formatted, Info}`, while the unit tests show otherwise.

I have no idea how/why dialyzer infers that. @michalmuskala, any clue?

awalterschulze: What I usually do is start adding specs down the call stack to try and debug dialyzer error.  They are cryptic and seem unreasonable sometimes, but they have also found a bug for me, at least.

PR #298, opened 2021-06-03
[Code health] Move CLI related tests to dedicated suite.

slaykachu:


awalterschulze:
So we need copies of the files that we use in both erlfmt and cli suites?

awalterschulze (approved):

slaykachu:
> So we need copies of the files that we use in both erlfmt and cli suites?

I knew you would spot that! Yes, since the data dir is based on the test suite name.
We *could* have used a common dir and play with the path, but I preferred to stick with regular CT structure.
* It's actually more robust: different tests have their own version of the file, which can be tweaked individually without impacting the other tests.
* We happened to reuse files, but this was a spurious factorisation in my book.
* Disc space is cheaper than development time!

PR #299, opened 2021-06-03
[Format range] Add --range as a command line option.

slaykachu:


src/erlfmt_cli.erl
```diff
@@ -361,24 +395,27 @@ resolve_config(
...
+        print_width = resolve_undefined(PreferWidth, DefaultWidth),
         pragma = resolve_pragma(PreferPragma, DefaultPragma),
-        out = resolve_out(PreferOut, DefaultOut)
+        out = resolve_out(PreferOut, DefaultOut),
+        range = resolve_undefined(PreferRange, DefaultRange)
     }.
 
-resolve_width(undefined, W) -> W;
-resolve_width(W, _) -> W.
+resolve_undefined(undefined, W) -> W;
```
awalterschulze, line 417: A nit, take it or leave it
But I preferred having verbose functions for each parameter

test/erlfmt_cli_SUITE.erl
```diff
@@ -236,6 +238,11 @@ exclude_check(Config) when is_list(Config) ->
     ?assertNotMatch(nomatch, string:find(WithoutBroken, "[warn]")),
     ?assertMatch(nomatch, string:find(WithoutBroken, "broken.erl")).
 
+range_check(Config) when is_list(Config) ->
```
awalterschulze, line 241: Nice <3

Is it possible to add one more with partial range?

I guess this would require making a more complicated stdio_test function, hmmm.
Maybe we can do this in a follow up commit, after this one is merged

src/erlfmt_cli.erl
```diff
@@ -311,6 +323,28 @@ parse_opts([delete_pragma | _Rest], _Files, _Exclude, #config{pragma = require})
     {error, "Cannot use both --require-pragma and --delete-pragma options together."};
 parse_opts([delete_pragma | Rest], Files, Exclude, Config) ->
     parse_opts(Rest, Files, Exclude, Config#config{pragma = delete});
+parse_opts([{range, String} | Rest], Files, Exclude, Config) ->
+    % Ad-hoc parsing. Mitigation for the absence of "couple of integers" direct support.
+    Range1 = string:split(String, ",", all),
+    Range2 = [list_to_integer(X) || X <- Range1],
+    Range3 =
+        case Range2 of
+            [X] ->
```
awalterschulze, line 332: I think is okay to not support a single line, except if this is required by the language server

src/erlfmt_cli.erl
```diff
@@ -156,21 +162,27 @@ unprotected_with_config(Name, ParsedConfig) ->
...
     case Verbose of
         true -> io:format(standard_error, "Formatting ~s\n", [FileName]);
         false -> ok
     end,
     Options =
         [{pragma, Pragma}] ++
             [{print_width, PrintWidth} || PrintWidth =/= undefined] ++
-            [verbose || Verbose],
+            [verbose || Verbose] ++
+            [{range, Range} || range =/= undefined],
```
awalterschulze, line 175: Nice

src/erlfmt.erl
```diff
@@ -234,7 +262,7 @@ format_file_range(FileName, StartLocation, EndLocation, Options) ->
     string(),
     erlfmt_scan:location(),
     erlfmt_scan:location(),
-    [{print_width, pos_integer()}]
+    config()
```
awalterschulze, line 265: The options `verbose` and `pragma` are not supported by this function, that is why this is a smaller property list.
Could we change this back?

src/erlfmt.erl
```diff
@@ -91,6 +107,18 @@ format_file(FileName, Options) ->
 -spec format_string(string(), config()) ->
     {ok, string(), [error_info()]} | {skip, string()} | {error, error_info()}.
 format_string(String, Options) ->
+    Range = proplists:get_value(range, Options),
+    case Range of
+        undefined ->
+            % Whole file (default).
+            format_string_full(String, Options);
```
awalterschulze, line 114: Nice

awalterschulze: I really like how well these functions are named and how easily this reads in both `format_string` and `format_file` functions.

src/erlfmt.erl
slaykachu: So, the range was extracted for dispatching purpose, but still sits in this properly list. Would I rather:
* Remove it from the properly list? Con: that's additional work just to 
* Introduce a sub_config() type with just `print_width` and `range`?
* Other. Con: I have no idea what this alternative is.

src/erlfmt_cli.erl
slaykachu: Well, supportting that just take this 2 simple lines. Does the right thing TM + 'Be liberal in what you accept' principle :)

src/erlfmt_cli.erl
slaykachu: Argl, it would trigger my DRY obsession!

test/erlfmt_cli_SUITE.erl
slaykachu: Exactly, that would be a separate commit, or even a separate issue (a new feature, whose benefit/use-case are unclear to me).

src/erlfmt.erl
awalterschulze: We want the function spec to only support what the function actually can support and not say that it can actually influence verbosity or where the output goes.

src/erlfmt_cli.erl
awalterschulze: I guess I would say being literal is that a range only supports a range, not a single line, but that is just my opinion.  Not going to block based on this.

src/erlfmt_cli.erl
awalterschulze: No problem, but just to explain where my review is coming from.
I prefer readability over DRY.
DRY is only important to me if it removes a large maintenance burden.

src/erlfmt.erl
slaykachu: Done :)

src/erlfmt_cli.erl
slaykachu: Having two functions doing the same thing doesn't match my definition of readability :)
'Same things should share a same name, different things should have distinct names'
Yet, I don't have strong feelings for this one, so I've switched back to one function per parameter.

awalterschulze (approved):

PR #300, opened 2021-06-07
[Forrmat range] Re-inject formatted range in original source.

slaykachu:


src/erlfmt.erl
```diff
@@ -272,7 +274,81 @@ format_file_range(FileName, StartLocation, EndLocation, Options) ->
 ) ->
     {ok, string(), [error_info()]}
     | {error, error_info()}.
-format_string_range(String, StartLocation, EndLocation, Options) ->
+format_string_range(Original, StartLocation, EndLocation, Options) ->
+    FileName = "nofile",
+    Pragma = proplists:get_value(pragma, Options, ignore),
+    {ok, Nodes, Warnings} = read_nodes_string(FileName, Original, Pragma),
```
michalmuskala, line 280: Wouldn't setting the `nofile` here result in bad diagnostics when called from the `format_file_range` function?

src/erlfmt.erl
```diff
@@ -639,3 +731,19 @@ check_line_lengths(FileName, Width, String, {FirstLineNo, _}) ->
...
+    Res =
+        lists:sublist(Target, Start - 1) ++
+            Injected ++
+            lists:sublist(Target, Start + Len, length(Target)),
+    if
+        length(Target) - Len + length(Injected) =/= length(Res) ->
+            error(assertion_error);
+        true ->
+            ok
+    end,
```
michalmuskala, line 748: FWIW assertions like that are commonly written with:
```
true = length(Target) - Len + length(Injected) =:= length(Res)
```

michalmuskala (commented):
Do we really need both functions? `format_file_range` and `format_file_range_extract`?

slaykachu:
> Do we really need both functions? format_file_range and format_file_range_extract?
No, but I'd rather keep some flexibility until the IDE integration is done, and then cleanup.

src/erlfmt.erl
slaykachu: Oups, good point, thanks!

awalterschulze:
Yeah I would also remove the function that we intend not to use and only test the one function.

test/erlfmt_SUITE.erl
```diff
@@ -1094,8 +1086,23 @@ snapshot_range_partial(_) ->
         "z()   ->  2.\n",
     % Only x() should be touched.
     Reference = "x() -> 0.\n",
-    Output = erlfmt:format_string_range(Original, {1, 1}, {1, 8}, []),
-    assert_snapshot_match(list_to_binary(Reference), Output).
+    Result = erlfmt:format_string_range_extract(Original, {1, 1}, {1, 8}, []),
+    assert_snapshot_match_range(Reference, Result).
+
+snapshot_range_partial_reinjected(_) ->
```
awalterschulze, line 1092: Would be nice to include a few more tests for reinjecting that formats the middle one, only the final one, none of them, all of them, two of them.

test/erlfmt_SUITE.erl
```diff
@@ -1147,11 +1187,29 @@ snapshot_enclosing_range_no_leak(_) ->
...
+        "\n"
+        "y()   ->  1.\n",
     % End point overshots line 3, but doesn't reach line 4:
     % Only x() must be formatted.
-    Output = erlfmt:format_string_range(Original, {1, 1}, {3, 3}, []),
-    assert_snapshot_match(list_to_binary(Reference), Output).
+    Result = erlfmt:format_string_range(Original, {1, 1}, {3, 3}, []),
+    assert_snapshot_match_range(Reference, Result).
+
+snapshot_range_reinjected(_) ->
```
awalterschulze, line 1198: Great test

src/erlfmt.erl
```diff
@@ -272,11 +272,48 @@ format_file_range(FileName, StartLocation, EndLocation, Options) ->
...
+    % Remove spurious empty line introduced by last \n separator.
+    FormattedAsList =
+        case lists:last(FormattedAsList0) of
+            <<>> ->
+                lists:droplast(FormattedAsList0);
+            _ ->
+                FormattedAsList0
+        end,
+    if
+        EndLine < StartLine ->
```
awalterschulze, line 309: Hmm good catch
I can't remember, do we maybe check this during argument parsing to give the user a nice error when this is not the case?

slaykachu: can do that! Will also use the idiomatic construct suggested by @michalmuskala elsewhere. 
Would give:
`true = EndLine < StartLine`

awalterschulze: Well when it is not `true` then the error given to the user would be stack trace instead of a nicely formatted error, or am I misunderstanding?

slaykachu: I forgot about that, but there is already an user-friendly check in `erlfmt_cli.erl`. It gives:
`Range: End must be greater or equal than Start`

awalterschulze: Great, then possibly we don't also need a check here?

slaykachu: Asserts don't hurt and have documentation value! It's also useful if the function get plugged from another client than the CLI.

test/erlfmt_SUITE.erl
slaykachu: Done!

test/erlfmt_SUITE.erl
```diff
@@ -1147,11 +1230,29 @@ snapshot_enclosing_range_no_leak(_) ->
...
+    % Verify the formatted range is properly re-injected
+    % in original file, even if the number of lines has changed.
+    Original =
+        "x()->0. % This comment end up in new line.\n"
```
awalterschulze, line 1245: Maybe we do want a few more tests where the number of lines are becoming more, like in this one.  But also where it is not the first form.
But I think it is okay to add that after merging this.

awalterschulze (approved):

test/erlfmt_SUITE.erl
slaykachu: I also think it is okay :).  This version is fit to test IDE integration!

PR #309, opened 2021-06-23
Introduce semi-expanded format for all containers

michalmuskala:
This change implements the semi-expanded format introduced for function calls in https://github.com/WhatsApp/erlfmt/pull/305 to all the containers.

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -415,7 +415,7 @@ with_next_break_fits(false, Doc, Fun) ->
     Fun(Doc).
 
 container(Meta, Values, Left, Right) ->
-    container_common(Meta, Values, Left, Right, break, last_normal).
+    container_common(Meta, Values, Left, Right, group_break, last_normal).
 
 flex_container(Meta, Values, Left, Right) ->
     container_common(Meta, Values, Left, Right, flex_break, last_fits).
@@ -461,12 +461,12 @@ break_behaviour(Meta, Values, BreakKind) ->
                 [SoleElement] ->
                     case is_container(SoleElement) of
                         true -> no_break;
-                        false -> break
+                        false -> group_break
                     end;
                 [First | _] ->
                     case is_tag(First) of
                         true -> BreakKind;
-                        false -> break
+                        false -> group_break
                     end;
                 _ ->
                     BreakKind
@@ -852,7 +852,7 @@ is_next_break_fits({FlexContainer, Meta, Values}) when
     FlexContainer =:= tuple; FlexContainer =:= bin
 ->
     BreakBehaviour = break_behaviour(Meta, Values, flex_break),
-    BreakBehaviour =:= break orelse BreakBehaviour =:= line;
+    BreakBehaviour =:= group_break orelse BreakBehaviour =:= line;
 is_next_break_fits(Expr) ->
     lists:member(element(1, Expr), [map, list, record, block, 'fun', lc, bc]) andalso
         has_no_comments_or_parens(Expr).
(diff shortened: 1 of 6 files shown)


src/erlfmt_format.erl
```diff
@@ -421,58 +421,55 @@ flex_container(Meta, Values, Left, Right) ->
...
+    container_common(Meta, Values, Left, Right, break, last_fits).
 
 container_common(_Meta, [], Left, Right, _, _) ->
     concat(Left, Right);
-container_common(Meta, Values, Left, Right, BreakKind0, LastFits) ->
+container_common(Meta, Values, Left, Right, BreakKind, LastFits) ->
     LastFitsFun = last_fits_fun(LastFits, Values),
-    BreakKind = break_behaviour(Meta, Values, BreakKind0),
-    BreakFun = break_fun(BreakKind),
+    BreakBehaviour = break_behaviour(Meta, Values, BreakKind),
```
awalterschulze, line 430: Nice naming

src/erlfmt_format.erl
```diff
@@ -535,16 +527,14 @@ has_opening_line_break(Meta, [HeadValue | _]) ->
 
 has_any_break_between([Head | [Head2 | _] = Tail]) ->
     has_break_between(Head, Head2) orelse has_any_break_between(Tail);
+has_any_break_between([{cons, _, Head, Tail}]) ->
```
awalterschulze, line 530: Ah I see this is for the new case you brought up

src/erlfmt_format.erl
```diff
@@ -491,18 +488,13 @@ is_tag(Expr) ->
...
-    surround(Left, <<"">>, concat(force_breaks(), group(Doc)), <<"">>, Right);
-surround_container(group_break, Left, Doc, Right) ->
-    surround(Left, <<"">>, group(Doc), <<"">>, Right);
 surround_container(no_break, Left, Doc, Right) ->
     concat(Left, Doc, Right);
 surround_container(flex_break, Left, Doc, Right) ->
-    group(concat(nest(concat(Left, Doc), ?INDENT, break), Right)).
+    group(concat(nest(concat(Left, Doc), ?INDENT, break), Right));
+surround_container(#break{inner = ForceInner, outer = ForceOuter}, Left, Doc, Right) ->
+    InnerDoc = group(concat(maybe_force_breaks(ForceInner), Doc)),
```
awalterschulze, line 496: So Doc is now always grouped, does this still cover the `line` case, which was not grouped?

test/erlfmt_format_SUITE.erl
```diff
@@ -3232,6 +3260,22 @@ exportimport(Config) when is_list(Config) ->
         "    c/1,\n"
         "    c/2\n"
         "]).\n"
+    ),
+    ?assertSame(
+        "-export([\n"
+        "    bar/1, baz/1\n"
```
awalterschulze, line 3266: Oh interesting.
I guess this is now a thing.
Almost feel like export should be special cased, but it is probably fine.

test/erlfmt_format_SUITE.erl
```diff
@@ -3232,6 +3260,22 @@ exportimport(Config) when is_list(Config) ->
...
+    ),
+    ?assertFormat(
+        "-export([\n"
+        "    foo/1, bar/1,\n"
+        "    baz/1, baz/2\n"
+        "]).\n",
+        "-export([\n"
+        "    foo/1,\n"
+        "    bar/1,\n"
+        "    baz/1, baz/2\n"
```
awalterschulze, line 3277: Great test

awalterschulze (approved):

src/erlfmt_format.erl
michalmuskala: Yes, using `force_breaks()` turns `break()` into `line()` so that still works correctly and removes one case

test/erlfmt_format_SUITE.erl
michalmuskala: Yes, I thought about this as well, but it seems consistent, if everything fits on a single line

PR #317, opened 2021-09-18
Support formatting PIDs, Refs and Ports

### LINKED ISSUE #316: Support formatting pids
It's useful when formatting some logs or other printed data. Also - Erlang supports pid literals in the shell, so this further makes sense.

Potentially we could also support formatting refs and ports for completeness.

For example:
```erlang
{foobar, <0.2769.0>}.
```

michalmuskala:
Closes #316

### DIFF
--- a/test/erlfmt_SUITE.erl
+++ b/test/erlfmt_SUITE.erl
@@ -49,6 +49,7 @@
     types/1,
     annos/1,
     shebang/1,
+    dotted/1,
     snapshot_simple_comments/1,
     snapshot_big_binary/1,
     snapshot_attributes/1,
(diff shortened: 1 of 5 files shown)


test/erlfmt_SUITE.erl
```diff
@@ -950,6 +952,20 @@ shebang(Config) when is_list(Config) ->
         )
     ).
 
+dotted(Config) when is_list(Config) ->
```
awalterschulze, line 955: Possibly I would add a failing test, that does not parse, but your choice

test/erlfmt_format_SUITE.erl
```diff
@@ -210,6 +212,11 @@ literals(Config) when is_list(Config) ->
     ?assertSame("_Bar\n"),
     ?assertFormat("$ ", "$\\s\n").
 
+dotted(Config) when is_list(Config) ->
```
awalterschulze, line 215: Very nice

awalterschulze (approved):

PR #318, opened 2021-09-30
Fix mistaken identity in description of CLI flag

frou:


michalmuskala:
Hah, good catch. Thanks! ❤️

PR #324, opened 2021-11-26
Improve error_info() type

gomoripeti:
`erlfmt_scan` returns `erl_anno:location()` in errors however `erlfmt_parse` mostly returns full `erlfmt_scan:anno()` maps. (I'm not sure if the parser should/can be changed to also return `location()` as the code generated by yecc just takes the second element of the token tuple, without any transformation, if it is not an `erl_anno:anno()` https://github.com/erlang/otp/blob/maint/lib/parsetools/include/yeccpre.hrl#L147) The errors from the parser also surface in the return value of `erlfmt:read_nodes[_string]`

I noticed this via dialyzer when calling `erlfmt_parse` directly.

### DIFF
--- a/src/erlfmt_parse.yrl
+++ b/src/erlfmt_parse.yrl
@@ -994,7 +994,7 @@ Erlang code.
 
 %% XXX. To be refined.
 -type error_description() :: term().
--type error_info() :: {erl_anno:line(), module(), error_description()}.
+-type error_info() :: {erl_anno:location() | erlfmt_scan:anno(), module(), error_description()}.
 -type token() :: erlfmt_scan:token().
 
 %% mkop(Op, Arg) -> {op,Anno,Op,Arg}.
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -34,7 +34,9 @@
 
 -export_type([error_info/0, config/0, pragma/0]).
 
--type error_info() :: {file:name_all(), erl_anno:location(), module(), Reason :: any()}.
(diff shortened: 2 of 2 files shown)


src/erlfmt_parse.yrl
```diff
@@ -994,7 +994,7 @@ Erlang code.
 
 %% XXX. To be refined.
 -type error_description() :: term().
--type error_info() :: {erl_anno:line(), module(), error_description()}.
+-type error_info() :: {erl_anno:location() | erlfmt_scan:anno(), module(), error_description()}.
```
gomoripeti, line 997: `ret_err/2` throws a proper `erl_anno:location()`, not sure if it can be returned by `parse_node` or if it always returns `erlfmt_scan:anno()`

gomoripeti:
Closing in favour of #325

gomoripeti:
without this fix I get dialyzer warnings when calling `erlfmt_parse:parse_node/1` directly
```
case erlfmt_parse:parse_node(Tokens) of
  {error, {ErrorLoc, erlfmt_parse, _Reason}} ->
    case ErrorLoc of
      #{'location' := Loc} -> ...
      Loc = {Line, Col} -> ...
```

```
The pattern #{'location':=Loc} can never match the type non_neg_integer()
The pattern Loc = {Line, Col} can never match the type non_neg_integer()
```

gomoripeti:
(hm, I intended to reopen when adding a comment)

PR #335, opened 2022-06-23
Ensure the correct IO opts when reading from stdin

filmor:
Since the `io:setopts` call is done in the escript `main`, it doesn't carry over to `rebar3 fmt` and breaks `rebar3 fmt -`. By adjusting the `standard_io` opts when actually reading from stdin and reverting back, this should work now.


michalmuskala:
Thank you! :heart:

PR #337, opened 2022-07-20
Allow erlfmt:format_string/2 to set filename from option

### LINKED ISSUE #336: Allow erlfmt:format_string to acccept filename as an option
Hi, 

I'm working on [erlfmt_formatter_plugin](https://github.com/wingyplus/erlfmt_formatter_plugin) which's bridge between mix formatter plugin and erlfmt. I found out that the `erlfmt:format_string/2` doesn't allow to set filename (https://github.com/WhatsApp/erlfmt/blob/main/src/erlfmt.erl#L132), this cause the formatter plugin display filename as `nofile` when combining with `erlfmt:format_error_info`. 

My proposal is allow `erlfmt:format_string/2` to accept filename as an options and default to `nofile` when it doesn't set. I can open a PR to implement this. 

What do you think?

wingyplus:
Closes #336

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -32,11 +32,12 @@
     format_string_range/4
 ]).
 
--export_type([error_info/0, config/0, pragma/0]).
+-export_type([error_info/0, config_option/0, config/0, pragma/0]).
 
 -type error_info() :: {file:name_all(), erl_anno:location(), module(), Reason :: any()}.
 -type pragma() :: require | insert | delete | ignore.
--type config() :: [{pragma, pragma()} | {print_width, pos_integer()} | verbose].
+-type config_option() :: {pragma, pragma()} | {print_width, pos_integer()} | verbose.
+-type config() :: [config_option()].
 
 -define(DEFAULT_WIDTH, 100).
 
@@ -108,7 +109,7 @@ format_file_full(FileName, Options) ->
         {error, Error} -> {error, Error}
     end.
 
--spec format_string(string(), config()) ->
+-spec format_string(string(), [config_option() | {filename, string()}]) ->
     {ok, string(), [error_info()]} | {skip, string()} | {error, error_info()}.
 format_string(String, Options) ->
     Range = proplists:get_value(range, Options),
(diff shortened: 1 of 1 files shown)


src/erlfmt.erl
```diff
@@ -36,7 +36,9 @@
 
 -type error_info() :: {file:name_all(), erl_anno:location(), module(), Reason :: any()}.
 -type pragma() :: require | insert | delete | ignore.
--type config() :: [{pragma, pragma()} | {print_width, pos_integer()} | verbose].
+-type config() :: [
+    {pragma, pragma()} | {print_width, pos_integer()} | {filename, string()} | verbose
+].
```
michalmuskala, line 41: The `config()` option is used in more places than just in `format_string`, and it seems in others we don't support the `{filename, _}` option. It might make sense to have a separate type there.

wingyplus: It means we should have a config type for format function?

wingyplus: Or we just move available config in the list to a new type and union it with filename config at format_string spec

michalmuskala: The second option is fine with me

wingyplus: Fixed. 🙇

src/erlfmt.erl
```diff
@@ -32,11 +32,12 @@
     format_string_range/4
 ]).
 
--export_type([error_info/0, config/0, pragma/0]).
+-export_type([error_info/0, config_option/0, config/0, pragma/0]).
 
 -type error_info() :: {file:name_all(), erl_anno:location(), module(), Reason :: any()}.
 -type pragma() :: require | insert | delete | ignore.
--type config() :: [{pragma, pragma()} | {print_width, pos_integer()} | verbose].
+-type config_option() :: {pragma, pragma()} | {print_width, pos_integer()} | verbose.
```
wingyplus, line 39: Type name can be change. 🙇

src/erlfmt.erl
```diff
@@ -108,7 +109,7 @@ format_file_full(FileName, Options) ->
         {error, Error} -> {error, Error}
     end.
 
--spec format_string(string(), config()) ->
+-spec format_string(string(), [config_option() | {filename, string()}]) ->
     {ok, string(), [error_info()]} | {skip, string()} | {error, error_info()}.
 format_string(String, Options) ->
     Range = proplists:get_value(range, Options),
```
wingyplus, line 115: I think we missing range in typespec. I can fix it in a new PR

michalmuskala:
Thank you! :heart:

PR #344, opened 2022-08-06
Fix markdown issues raised by markdownlint

kianmeng:
Ran against this command:

    markdownlint *.md doc/*.md --disable MD013 MD034 MD010 MD051 MD024 MD029

See a detailed description of the rules is available at
https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md


RELEASE.md
```diff
@@ -3,50 +3,58 @@
...
+Description: Copy ChangeLog contents
 ```
+
 6. Release to WhatsApp.
    This could include building with an older version of erlang.
    Also update the `rebar.config.script` with new version tag and reformat code base.
    Here is an example of building with erlang version 22 on mac:
+
 ```bash
-$ brew install erlang@22
```
awalterschulze, line 31: Here the dollars vs hashtags are kind of important

RELEASE.md
```diff
@@ -3,50 +3,58 @@
...
+_build/release/bin/erlfmt -h
 ```
+
 7. Release to hex: `rebar3 hex publish`.
 
 If you have not used hex before, create `~/.config/rebar3/rebar.config` and include the contents `{plugins, [rebar3_hex]}`:
+
 ```bash
-$ mkdir -p ~/.config/rebar3/ && echo "{plugins, [rebar3_hex]}." >> ~/.config/rebar3/rebar.config
+mkdir -p ~/.config/rebar3/ && echo "{plugins, [rebar3_hex]}." >> ~/.config/rebar3/rebar.config
```
awalterschulze, line 50: Same here the dollar indicates the command to run and the non dollar indicates the output

StyleGuide.md
```diff
@@ -88,48 +104,58 @@ read_file
...
-  ```
+
+```erlang
+{elvis_style, module_naming_convention, #{regex => "^([a-z][a-z0-9]*_?)*(_SUITE)?$"}}
+```
 
 ### Booleans
 
-* avoid `and` and `or` operators, prefer` andalso` and `orelse`, or
+* avoid `and` and `or` operators, prefer `andalso` and `orelse`, or
```
awalterschulze, line 156: Nice catch here

awalterschulze (commented):
In general I am against removing the `$` and `#` from verbatim text.  Otherwise it looks like an improvement.  Pretty nice tool :)

RELEASE.md
kianmeng:
```
RELEASE.md:52:1 MD014/commands-show-output Dollar signs used before commands without showing output [Context: "$ mkdir -p ~/.config/rebar3/ &..."]
```

RELEASE.md
kianmeng: Agree, hashtags are comments, revert changes.

awalterschulze:
I would keep the `$`s everywhere, but otherwise I would accept.

If @michalmuskala thinks removing the `$`s are fine, then I don't object to merging.

kianmeng:
@michalmuskala Any updates on this PR?

PR #350, opened 2023-09-10
Allow macro as case clause.

ztion:
If you run erlfmt on a file like this:

```
-module(testar).

-define(tjo(A), A -> ok).

apa(B) ->
      case B of
   ?tjo(1)
   ;
          ?tjo(2)
	end.
```

Then you will get this error
testar.erl:8:4: syntax error before: ';'
and the apa function will not be formatted. (There is another error also for the define, but that is not solved by the current PR)

The parse error is due to the parser not allowing a macro as the whole case clause and thus gets a parse error when getting to the semi-colon. This kind of macro usage is found in the code base I work with. With this PR, the apa function can be parsed and is successfully formatted.


michalmuskala:
Thank you for the PR. This looks like a sensible change, would you mind adding some tests to show things work properly as expected?

ztion:
Have updated PR with a test now!

michalmuskala:
Thank you! :heart:

PR #351, opened 2023-09-25
Implement `maybe_expr` support

### LINKED ISSUE #348: Syntax error on maybe_expr
I have the following module:

```erl
-module(erlfmt_maybe_app).
-feature(maybe_expr, enable).
-behaviour(application).
-export([start/2, stop/1]).

start(_StartType, _StartArgs) ->
maybe erlfmt_maybe_sup:start_link() end.

stop(_State) ->
    ok.
```

That compiles as expected with `erlc` while using Erlang/OTP 26.0.

But running `rebar3 fmt` gives the following output and the file is not formatted:
```
ᐅ DEBUG=1 rebar3 fmt
===> 26.0 satisfies the requirement for minimum OTP version 21
===> Compile (apps)
===> Expanded command sequence to be run: [fmt]
===> Running provider: fmt
src/erlfmt_maybe_app.erl:7:7: syntax error before: erlfmt_maybe_sup
```

Steps to reproduce:
```
ᐅ rebar3 new app erlfmt_maybe
ᐅ cd erlfmt_maybe
ᐅ echo "{project_plugins, [erlfmt]}." >> rebar.config
ᐅ echo "{erlfmt, [write]}." >> rebar.config
ᐅ vi src/erlfmt_maybe_app.erl # Edit the file to match the example above
ᐅ DEBUG=1 rebar3 fmt
===> 26.0 satisfies the requirement for minimum OTP version 21
===> Compile (apps)
===> Expanded command sequence to be run: [fmt]
===> Running provider: fmt
src/erlfmt_maybe_app.erl:7:7: syntax error before: erlfmt_maybe_sup

ᐅ rebar3 --version
rebar 3.22.0 on Erlang/OTP 26 Erts 14.0
ᐅ grep version _build/default/plugins/erlfmt/hex_metadata.config
{<<"version">>,<<"1.1.0">>}.
```

maennchen:
Resolves #348

I would've preferred to pass a compiler option to `erl_scan:tokens/4` or something similar, but that does not seem to exist. Therefore this change currently requires the environment variable `ERL_AFLAGS="-enable-feature all"` to be set. I'd love to hear a better solution.

If the environment variable is set, various other things break. This PR addresses them es well.

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -196,6 +196,18 @@ do_expr_to_algebra({args, Meta, Values}) ->
 do_expr_to_algebra({'case', _Meta, Expr, Clauses}) ->
     Prefix = surround(<<"case">>, <<" ">>, expr_to_algebra(Expr), <<" ">>, <<"of">>),
     surround_block(Prefix, clauses_to_algebra(Clauses), <<"end">>);
+do_expr_to_algebra({'maybe', _Meta, MaybeExpressions}) ->
+    surround_block(<<"maybe">>, maybe_expressions_to_algebra(MaybeExpressions), <<"end">>);
+do_expr_to_algebra({'maybe', _Meta, MaybeExpressions, Else}) ->
+    ElseD = expr_to_algebra(Else),
+    surround_block(
+        <<"maybe">>, maybe_expressions_to_algebra(MaybeExpressions), line(ElseD, <<"end">>)
+    );
+do_expr_to_algebra({else_clause, _Meta, Clauses}) ->
+    concat(
+        force_breaks(),
+        group(concat(<<"else">>, nest(concat(line(), clauses_to_algebra(Clauses)), ?INDENT)))
+    );
 do_expr_to_algebra({'receive', _Meta, Clauses}) ->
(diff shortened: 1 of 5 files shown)


michalmuskala:
Thanks, this is great!

I'll take some time to think how we could avoid passing in the flag - this definitely doesn't seem ideal

maennchen:
@michalmuskala Yeah, especially on codebases that already use the else reserved keywords in some modules and maybe in others.

michalmuskala:
This was manually merged in e8d6376fc079cec63095742749b8c4e83975157c.
This will be released in 1.3.0.

I made it so the code behaves as if the feature is always enabled.
At the same time, if disabling the feature would make the code parse, it will be re-formatted with the `maybe` and `else` atoms re-quoted.
In other worlds, erlfmt should always work - both on the code with feature enabled & disabled. On code with the feature disabled the new version will always format the `maybe` and `else` atoms as `'maybe'` and `'else'`.

maennchen:
@michalmuskala Thanks for merging and releasing. 

I've updated in `oidcc` and can verify that everything seems to behave as expected: https://github.com/erlef/oidcc/pull/284

PR #352, opened 2023-12-21
Improve error_info() type

gomoripeti:
`erlfmt_scan` returns `erl_anno:location()` in errors however `erlfmt_parse` mostly returns full `erlfmt_scan:anno()` maps. The errors from the parser also surface in the return value of `erlfmt:read_nodes[_string]`

I noticed this via dialyzer when calling erlfmt_parse directly.


michalmuskala:
Sorry for the big delay in merging this.

Thank you! ❤️

PR #356, opened 2024-08-17
Apply `exclude_files` to `files` absolute paths

danielfinke:
When calling `erlfmt` with absolute paths for the files to format, the excludes don't apply, even if the resolved paths of the excludes would match the files to format.

Resolve the paths of the excludes based on the current working directory and add them to the excludes.

This issue was observed when using the
[VS Code Erlang Formatter extension](https://marketplace.visualstudio.com/items?itemName=szTheory.erlang-formatter). The extension invokes `rebar3 fmt` with the full path of the file to format.

### DIFF
--- a/src/erlfmt_cli.erl
+++ b/src/erlfmt_cli.erl
@@ -107,8 +107,25 @@ with_parsed(Name, Config) ->
             erlang:halt(127)
     end.
 
+-spec set_difference([file:name_all()], [file:name_all()]) -> [file:name_all()].
 set_difference(Files, Excludes) ->
-    sets:to_list(sets:subtract(sets:from_list(Files), sets:from_list(Excludes))).
+    {ok, Dir} = file:get_cwd(),
+    AbsoluteExcludes = [resolve_path(Dir, E) || E <- Excludes],
+    AllExcludes = sets:from_list(Excludes ++ AbsoluteExcludes),
+    sets:to_list(sets:subtract(sets:from_list(Files), AllExcludes)).
+
+resolve_path(Dir, Filename) ->
+    case filename:pathtype(Filename) of
+        absolute -> Filename;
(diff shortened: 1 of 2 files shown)


awalterschulze (approved):
Looks good to me, what do you think @michalmuskala ?

michalmuskala (changes requested):
One concern about windows support.

The second concerns would be with symlinks. I'm not sure it would all work correctly in this case.

Wouldn't an alternative of rather than checking for file equality to check for prefixes work?

danielfinke:
@awalterschulze @michalmuskala ready for further review 🙂 

I've addressed Windows support by switching to using `filename:absname/2` for abspath resolution.

The change also makes sure `files`/`exclude_files` consistently resolve relative paths in case of double dots/parent directory segments. Note that previously, erlfmt didn't resolve these - e.g. `a/../b` and `b` were distinct. I had already resolved these for the absolute excludes in the first commit. Now it applies to all paths for both `files`/`exclude_files`.

I intentionally avoided any symlink resolution. The existing behaviour of erlfmt treats symlinks as distinct files. e.g.
```bash
$ echo "   -module(a)." > a.erl
$ ln -s a.erl b.erl
$ rebar3 fmt --verbose --exclude-files "a.erl" b.erl
Formatting b.erl
-module(a).
```
`a.erl` is skipped, because the `b.erl` symlink is not followed.

I think this makes sense and is consistent with most treatment of symlinks in my experience. One example is the script name argument when running a bash script:
```bash
$ echo 'echo $0' > test.sh
$ chmod +x test.sh
$ ln -s test.sh test2.sh
$ ./test.sh
./test.sh
$ ./test2.sh
./test2.sh
```
Unless the script explicitly follows its script name symlink, it thinks it is a unique script.

michalmuskala:
Thank you! ❤️

PR #358, opened 2024-08-24
Support for OTP 27 Sigils & Triple Quoted Strings

### LINKED ISSUE #355: OPT 27's `-moduledoc """` makes `erlfmt` fail
Failure is something like `_build/default/plugins/erlfmt/src/erlfmt_parse.yrl:98:71: Warning: expression updates a literal`

I'm guessing you're aware of the issue, by now, but this can serve as a way to follow-up on when a fix is coming. 👍 

Thanks.

maennchen:
Sigils: https://www.erlang.org/blog/highlights-otp-27/#sigils
Triple Quoted Strings: https://www.erlang.org/blog/highlights-otp-27/#triple-quoted-strings

Also adds tests for documentation attributes. See: https://www.erlang.org/blog/highlights-otp-27/#overhauled-documentation-system

Resolves #355

### DIFF
--- a/src/erlfmt_parse.yrl
+++ b/src/erlfmt_parse.yrl
@@ -608,7 +608,7 @@ atomic -> string : '$1'.
 atomic -> string concatables : {concat, ?range_anno('$1', '$2'), ['$1' | ?val('$2')]}.
 atomic -> macro_call_none concatables : {concat, ?range_anno('$1', '$2'), ['$1' | ?val('$2')]}.
 atomic -> macro_string concatables : {concat, ?range_anno('$1', '$2'), ['$1' | ?val('$2')]}.
-atomic -> sigil_prefix string sigil_suffix : {sigil, ?range_anno('$1', '$3'), {'$1', '$2', '$3'}}.
+atomic -> sigil_prefix string sigil_suffix : {sigil, ?range_anno('$1', '$3'), '$1', '$2', '$3'}.
 
 concatables_no_initial_call -> concatable_no_call concatables : {['$1' | ?val('$2')], ?anno('$2')}.
 concatables_no_initial_call -> concatable_no_call : {['$1'], ?anno('$1')}.
--- a/src/erlfmt_parse.yrl
+++ b/src/erlfmt_parse.yrl
@@ -62,7 +62,8 @@ char integer float atom string var
 '<<' '>>'
 '?' '!' '=' '::' '..' '...'
 spec callback define_expr define_type define_clause standalone_exprs % helper
-dot.
+dot
+sigil_prefix sigil_suffix.
 
 %% Conflict comes from optional parens on macro calls.
 Expect 1.
@@ -607,6 +608,7 @@ atomic -> string : '$1'.
 atomic -> string concatables : {concat, ?range_anno('$1', '$2'), ['$1' | ?val('$2')]}.
 atomic -> macro_call_none concatables : {concat, ?range_anno('$1', '$2'), ['$1' | ?val('$2')]}.
 atomic -> macro_string concatables : {concat, ?range_anno('$1', '$2'), ['$1' | ?val('$2')]}.
+atomic -> sigil_prefix string sigil_suffix : {sigil, ?range_anno('$1', '$3'), {'$1', '$2', '$3'}}.
 
 concatables_no_initial_call -> concatable_no_call concatables : {['$1' | ?val('$2')], ?anno('$2')}.
 concatables_no_initial_call -> concatable_no_call : {['$1'], ?anno('$1')}.
--- a/src/erlfmt_scan.erl
+++ b/src/erlfmt_scan.erl
@@ -235,14 +235,7 @@ drop_initial_white_space(Rest) ->
 
 -spec stringify_tokens([erl_scan:token()]) -> string().
 stringify_tokens(Tokens) ->
-    lists:flatmap(fun stringify_token/1, Tokens).
-
--spec stringify_token(erl_scan:token()) -> string().
-stringify_token(Token) ->
-    case erl_scan:text(Token) of
-        undefined -> "";
-        Text -> Text
-    end.
+    lists:flatmap(fun erl_scan:text/1, Tokens).
 
 -spec split_tokens([erl_scan:token()], [erl_scan:token()]) ->
     {[token()], [erl_scan:token()], [comment()], [token()]}.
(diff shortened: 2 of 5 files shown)


eproxus:
Lovely! I had just started on a similar PR myself 😄 

Does it support triple quoted strings in documentation attributes? I didn't see a specific test case about that:

```erlang
-module(foo).
-moduledoc """
The module.
""".

-doc """
The function.

Long description.
""".
f() -> ok.
```

Also, does it re-indent triple quoted strings. E.g. from:

```erlang
f() ->
    """
the
  long
     string
""".
```
to
```erlang
f() -> 
    """
    the
      long
         string
    """.
```


I used the following test module `test/erlfmt_SUITE_data/triple_quoted_strings.erl`:
```erlang
-module(triple_quoted_strings).
-moduledoc """
abc
""".

-doc "foobar".
-type my_type :: integer().

-doc """
     baz
         extra
     """.
-opaque otype :: integer().

-doc "qux"
foo() ->
    """
    foo
    """.

bar() ->
"""
the
  long
     string
""".
```
Together with the desired result `test/erlfmt_SUITE_data/triple_quoted_string.erl.formatted`:
```erlang
-module(triple_quoted_strings).
-moduledoc """
abc
""".

-doc "foobar".
-type my_type :: integer().

-doc """
     baz
         extra
     """.
-opaque otype :: integer().

-doc "qux"
foo() ->
    """
    foo
    """.

bar() ->
    """
    the 
      long
         string
    """.
```
as a test driver, though I'm not 100% sure about if the result will be correct since I never finished implementing the indentation logic.

maennchen:
> Does it support triple quoted strings in documentation attributes? I didn't see a specific test case about that:

Yes.

https://github.com/WhatsApp/erlfmt/pull/358/files#diff-ca49d93bca04b364c997af828c3aaf95beff334a9307b97fb1317b1721812ae4R4271

> Also, does it re-indent triple quoted strings. E.g. from:

No, I haven't looked at indentation. But that would be a good feature to have.

I don't have the time right now to also get into this myself, but I would be very happy to collaborate to create a shared PR combined with your indentation logic. Do you want to send a PR to my PR branch? :D

michalmuskala:
Thank you for the PR, this is great!

I agree we do need to handle indentation of the tripple-quote strings before merging. The test example that @eproxus gave is a good one to include, however I see an inconsistency with:
```erl
-moduledoc """
abc
""".

-doc """
     baz
         extra
     """.
```
I believe both should end up indented like the `-moduledoc` to end up with:
```erl
-moduledoc """
abc
""".

-doc """
baz
    extra
""".
```

As to the particular approach, I'm generally fine with a hacky solution - the formatter has many of those already 😅

src/erlfmt_parse.yrl
```diff
@@ -607,6 +608,7 @@ atomic -> string : '$1'.
 atomic -> string concatables : {concat, ?range_anno('$1', '$2'), ['$1' | ?val('$2')]}.
 atomic -> macro_call_none concatables : {concat, ?range_anno('$1', '$2'), ['$1' | ?val('$2')]}.
 atomic -> macro_string concatables : {concat, ?range_anno('$1', '$2'), ['$1' | ?val('$2')]}.
+atomic -> sigil_prefix string sigil_suffix : {sigil, ?range_anno('$1', '$3'), {'$1', '$2', '$3'}}.
```
michalmuskala, line 611: We can skip the inner tuple here
```suggestion
atomic -> sigil_prefix string sigil_suffix : {sigil, ?range_anno('$1', '$3'), '$1', '$2', '$3'}.
```

src/erlfmt_scan.erl
```diff
@@ -235,7 +235,14 @@ drop_initial_white_space(Rest) ->
 
 -spec stringify_tokens([erl_scan:token()]) -> string().
 stringify_tokens(Tokens) ->
-    lists:flatmap(fun erl_scan:text/1, Tokens).
+    lists:flatmap(fun stringify_token/1, Tokens).
+
+-spec stringify_token(erl_scan:token()) -> string().
+stringify_token(Token) ->
+    case erl_scan:text(Token) of
+        undefined -> "";
```
michalmuskala, line 243: In what case does this happen? This is used when skipping formatting and I'm afraid it might end up with some inconsistencies in this case.
I'd like to understand when `undefined` is a case here so we can properly test it

eproxus:
@michalmuskala I added the second in the sense of that the tool should try to keep the existing desire of the user if possible (as is done with line breaks somewhat). But perhaps for indentation that rule is not really applied now that I think about it...

test/erlfmt_format_SUITE.erl
```diff
@@ -232,6 +238,30 @@ literals(Config) when is_list(Config) ->
     ?assertSame("_Bar\n"),
     ?assertFormat("$ ", "$\\s\n").
 
+-if(?OTP_RELEASE >= 27).
```
michalmuskala, line 241: I don't think this needs to be a compile-time check. A runtime check would be enough since it doesn't use any of the new features inline.

I would suggest adding a `otp_27_features` group and then have an `init_per_group` callback with the check:
```erl
init_per_group(otp_27_features, Config) ->
    case erlang:system_info(otp_release) >= "27" of
        true -> Config;
        false -> {skip, "Skipping tests for features from OTP >= 27"}
    end;
init_per_group(_, Config) ->
    Config.
```

michalmuskala:
@eproxus in general we retain line breaks introduced by the user, but we do re-indent everything according to the rules

eproxus:
@michalmuskala Makes sense!

@maennchen Do you want a PR with my stuff right now to your branch, or would you rather fix the comments first?

src/erlfmt_scan.erl
maennchen: Hm, I though this was triggered by the sigil suffix (when empty). But it seems by implementing the sigils correctly, they no longer cause it to produce `undefined`. I'll remove the change.

maennchen:
@eproxus Fixes are done. Go ahead :)

michalmuskala:
Is this ready to merge? Or should I wait for the work on indentation?

eproxus:
I think the PR is good enough to add compatibility for OTP 27. I got stuck in the rabbit whole of actually implementing indentation and didn't succeed yet. @maennchen Do you want my half-working code there?

maennchen:
@michalmuskala We can merge this without indentation support. 👍

@eproxus I don't plan on working on the indentation. But if anyone else wants to, it might be good to push the code somewhere so that they can look at the work you already did.

michalmuskala:
I'll look into properly handling indentation myself

Thank you both for working on this! ❤️

michalmuskala:
Support for indentation was implemented in https://github.com/WhatsApp/erlfmt/pull/362 and I cut a 1.5.0 release with support for the new OTP 27 features

PR #359, opened 2024-08-28
Allow providing an optional reason when using erlfmt-ignore

jcpetruzza:
Right now only trailing spaces and `%` are allowed in an `erlfmt-ignore`. We make it so that anything after the `erlfmt-ignore` is discarded, effectively allowing for the user to document why formatting is disabled


michalmuskala:
Thank you! ❤️

PR #360, opened 2024-08-28
Allow erlfmt:ignore to be used in addition to erlfmt-ignore

jcpetruzza:
This puts erlfmt in line with other tools that use the format `toolname:ignore` to disable warnings, and makes it easier for developers to "guess" the correct incantation


michalmuskala:
Thank you! ❤️

PR #365, opened 2024-12-06
Set escript main parameter

dmitrivereshchagin:
If this parameter is not specified explicitly, erlfmt cannot be installed under a non-standard name:

    % cp _build/release/bin/erlfmt erlfmt-1.5.0
    % ./erlfmt-1.5.0
    escript: exception error: undefined function 'erlfmt-1.5':main/1
    ...

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -8,7 +8,7 @@
 
 {escript_name, erlfmt}.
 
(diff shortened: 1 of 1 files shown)


michalmuskala:
Thank you! ❤️

PR #366, opened 2025-01-29
Support trailing commas and semis

michalmuskala:
They will be just removed when formatting, allowing more "sloppy" code before reformatting, increasing usability of erlfmt.


elbrujohalcon:
I don't think I can persuade you folks from discarding this idea, but FWIW… It doesn't seem like a great one to me as it might just hide issues (particularly missing or misplaced clauses or expressions in a function). Is there any possibility that you would consider using a configuration parameter to enable/disable this behaviour? I don't mind if it's in "patch sloppy code" mode as a default, but I would very much like to be able to set it to "I can fix my syntax errors myself, thank you very much" mode.

elbrujohalcon:
Woah… took me too much to write that comment, I see… 🤷🏻

PR #367, opened 2025-04-25
Fix bug where two contiguous ignore-begin blocks results in an error

ruippeixotog:
Erlfmt currently can't parse two contiguous `erlfmt:ignore-begin` / `erlfmt:ignore-end` blocks. When we parse the first `ignore-end` we leave the formatter in an "end" state and don't clear it before the next "begin" since we don't run `ignore_state_pre` or `ignore_state_post` until the next non-comment. erlfmt doesn't handle transitions from `end` to `begin`, resulting in an error.

After checking the code, it's unclear to me why we need an `end` state in the first place - we always end up transforming it in `false` whenever we start or finish processing a statement. On this PR I'm simplifying things by simply setting the state to `false` after an `erlfmt:ignore-end` statement.

The unit test added fails in master and can be used to confirm the fix.

### DIFF
--- a/src/erlfmt.erl
+++ b/src/erlfmt.erl
@@ -538,15 +538,11 @@ parse(Tokens) ->
     end.
 
 ignore_state_pre(PreComments, FileName, Acc) ->
-    case ignore_state(PreComments, FileName, Acc) of
-        'end' -> false;
-        Other -> Other
-    end.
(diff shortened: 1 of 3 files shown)


michalmuskala:
Thank you! ❤️

The CI failing is due to ubuntu 20.04 deprecation, I'll fix that separately

PR #371, opened 2025-08-30
Fix sigil formatting bug that produced invalid mixed syntax

### LINKED ISSUE #369: Error when formatting sigils containing newlines
When trying to format code containing multiline sigils:
```
-module(test).

-export([hello/0]).

hello() ->
    ~"Hello
             there".
```
An error is reported:
```
test.erl:5:1: formatter result not equivalent. Please report this bug.
```
Hence I'm reporting the error - not sure how popular this sigil usage is though 😅

However, it's legal Erlang code. It compiles without warnings, and the sigil is properly handled:
```
1> test:hello().
<<"Hello\n             there">>
```

williamthome:
Fixes #369

When formatting multiline sigil strings, erlfmt was incorrectly applying regular string transformation logic, creating invalid mixed syntax like:

```erlang
~"\n"
"    foo\n"
"    "
```

This fix leaves the sigil content entirely unchanged to prevent the mixed syntax bug.

I'm not entirely sure if this is the best solution - sigils might still be able to participate in container line-breaking while keeping their raw content intact.

I welcome maintainer feedback on whether full preservation fits erlfmt's formatting philosophy or if an alternative approach would be better.

### DIFF
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -271,12 +271,6 @@ do_expr_to_algebra({sigil, _Meta, Prefix, {string, ContentMeta, _Value}, Suffix}
         % Multi-line non-triple-quoted sigils need force_breaks
         _ -> concat([force_breaks(), SigilDoc])
     end;
-do_expr_to_algebra({sigil, _Meta, Prefix, {Atomic, ContentMeta, _Value}, Suffix}) when
-    ?IS_ATOMIC(Atomic)
-->
-    PrefixDoc = concat(<<"~">>, do_expr_to_algebra(Prefix)),
-    Text = erlfmt_scan:get_anno(text, ContentMeta),
-    concat(concat(PrefixDoc, string(Text)), do_expr_to_algebra(Suffix));
 do_expr_to_algebra({sigil_prefix, _Meta, ''}) ->
     <<"">>;
 do_expr_to_algebra({sigil_prefix, _Meta, SigilName}) ->
--- a/src/erlfmt_format.erl
+++ b/src/erlfmt_format.erl
@@ -259,9 +259,12 @@ do_expr_to_algebra({clauses, _Meta, Clauses}) ->
     clauses_to_algebra(Clauses);
 do_expr_to_algebra({body, _Meta, Exprs}) ->
     block_to_algebra(Exprs);
-do_expr_to_algebra({sigil, _Meta, Prefix, Content, Suffix}) ->
+do_expr_to_algebra({sigil, _Meta, Prefix, {Atomic, ContentMeta, _Value}, Suffix}) when
+    ?IS_ATOMIC(Atomic)
+->
     PrefixDoc = concat(<<"~">>, do_expr_to_algebra(Prefix)),
-    concat(concat(PrefixDoc, do_expr_to_algebra(Content)), do_expr_to_algebra(Suffix));
+    Text = erlfmt_scan:get_anno(text, ContentMeta),
+    concat(concat(PrefixDoc, string(Text)), do_expr_to_algebra(Suffix));
 do_expr_to_algebra({sigil_prefix, _Meta, ''}) ->
     <<"">>;
 do_expr_to_algebra({sigil_prefix, _Meta, SigilName}) ->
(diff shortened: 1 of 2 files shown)


michalmuskala:
Sorry for the delay in reviewing this.

I'm fairly convinced this won't work correctly to indent the resulting strings inside sigils - the formatting algebra doesn't expect newlines inside strings - it expects to manage all newlines explicitly.

michalmuskala:
The only thing I'm still a bit apprehensive about is triple-quote strings - since then it's easy to properly handle newlines (and we already do it), I think for sigils with a triple-quote string inside we should do the previous code. They're not problematic since they strip-out the indentation whitespace properly. You can check out the snapshot tests in the `tripple_string.erl` - I think if we add a sigil to each one, they should behave the same as without it.

src/erlfmt_format.erl
```diff
@@ -259,9 +259,24 @@ do_expr_to_algebra({clauses, _Meta, Clauses}) ->
...
+    SigilDoc = concat(concat(PrefixDoc, string(Text)), do_expr_to_algebra(Suffix)),
+    case string:split(Text, "\n", all) of
+        % Triple-quoted sigils don't need force_breaks
+        ["\"\"\"" ++ _ | _] -> SigilDoc;
+        % Single-line sigils don't need force_breaks
+        [_] -> SigilDoc;
+        % Multi-line non-triple-quoted sigils need force_breaks
+        _ -> concat([force_breaks(), SigilDoc])
+    end;
+do_expr_to_algebra({sigil, _Meta, Prefix, {Atomic, ContentMeta, _Value}, Suffix}) when
```
michalmuskala, line 274: Per the parser, sigil can only have string inside. This clause will never be executed.

https://github.com/WhatsApp/erlfmt/blob/1292f7c96b0d055f2e4c1facfdb54802e2ebdfbb/src/erlfmt_parse.yrl#L653

src/erlfmt_format.erl
```diff
@@ -259,9 +259,24 @@ do_expr_to_algebra({clauses, _Meta, Clauses}) ->
...
     block_to_algebra(Exprs);
-do_expr_to_algebra({sigil, _Meta, Prefix, Content, Suffix}) ->
+do_expr_to_algebra({sigil, _Meta, Prefix, {string, ContentMeta, _Value}, Suffix}) ->
     PrefixDoc = concat(<<"~">>, do_expr_to_algebra(Prefix)),
-    concat(concat(PrefixDoc, do_expr_to_algebra(Content)), do_expr_to_algebra(Suffix));
+    Text = erlfmt_scan:get_anno(text, ContentMeta),
+    SigilDoc = concat(concat(PrefixDoc, string(Text)), do_expr_to_algebra(Suffix)),
+    case string:split(Text, "\n", all) of
+        % Triple-quoted sigils don't need force_breaks
+        ["\"\"\"" ++ _ | _] -> SigilDoc;
```
michalmuskala, line 268: This will not work correctly for multiple nesting levels. We need to do the same logic as is done in normal triple-quoted strings in `string_to_algebra`, just add prefix and suffix around it.
Furthermore, ideally we'd avoid duplicating this logic and just share the code.

test/erlfmt_format_SUITE.erl
```diff
@@ -295,7 +295,126 @@ sigils(Config) when is_list(Config) ->
     ),
     ?assertSame("~s\"\"\"\n\\tabc\n\\tdef\n\"\"\"\n"),
     ?assertSame("\"\"\"\nTest\nMultiline\n\"\"\"\n"),
-    ?assertSame("~\"\"\"\nTest\nMultiline\n\"\"\"\n").
+    ?assertSame("~\"\"\"\nTest\nMultiline\n\"\"\"\n"),
+    %% Test multiline sigils in lists cause list to break
+    ?assertFormat(
+        "[~\"\n    foo\n    \"].\n",
```
michalmuskala, line 301: For tests, please follow the existing code style and avoid `\n` directly in strings - split them across multiple lines. This makes it possible to easier understand what's going on from looking at the test - this is very hard to understand how things are moved around.

src/erlfmt_format.erl
williamthome: [08de55a](https://github.com/WhatsApp/erlfmt/pull/371/commits/08de55a5354dbddda7aeb754ac9363e7b979b0e7)

test/erlfmt_format_SUITE.erl
williamthome: [d8cbea4](https://github.com/WhatsApp/erlfmt/pull/371/commits/d8cbea4f40e9f14148dcce4d08b010ef75f1e637)

michalmuskala:
Alternative solution in https://github.com/WhatsApp/erlfmt/commit/5c1dc08573d44c74a507cd9912a65a0c9edb14bd that properly handles how we split things into newlines or not with just a final trailing newline

